PR #148, opened 2026-04-17
fix: Fixes cookie authentication and port vsp --browser-auth to TypeScript

samibouge:
## Summary

Fixes cookie authentication for both the CLI and the MCP server, surfaces verbose logs from the CLI, and hardens ADT requests against SAML2 login-page redirects.

Add standalone script src/browser-extract-cookie-auth.ts that launches a Chromium-based browser, captures SAP session cookies via CDP, and writes them in Netscape format (compatible with --cookie-file). Not wired into the main CLI because the code is ugly.

## Changes

- **Wire `SAP_COOKIE_FILE` / `SAP_COOKIE_STRING` through to the ADT client.** The env vars were parsed into `ServerConfig` but never forwarded to `AdtClient`, so cookies were silently dropped. Added a `resolveCookies()` helper in `src/adt/cookies.ts` and consumed it from `src/cli.ts` (`createClientFromEnv`) and `src/server/server.ts` (`buildAdtConfig`, which also feeds the per-user principal-propagation clients).

- **Initialize the logger in the CLI.** `SAP_VERBOSE` / `--verbose` had no effect on `npm run cli search/source` because `initLogger()` only ran inside `createAndStartServer()`. The CLI now calls `initLogger(logFormat, verbose)` in `createClientFromEnv()`, so debug-level `[http_request]` audit events show up on stderr.

- **Disable SAML2 on ADT requests.** SAML2-configured systems respond to unauthenticated requests with an HTML login page and HTTP 200, which breaks programmatic clients (ARC-1 previously returned the login HTML instead of a proper 401). `buildUrl()` now appends `saml2=disabled` so ICF falls back to Basic / cookie auth and returns real status codes.


marianfoo:
Thank you for your PR, was implemented with #149 please test, if something is not working, create a Issue

PR #173, opened 2026-04-21
feat: close RAP on-prem authoring gaps with preflight and handler scaffolding

marianfoo:
## Goal
Implement the RAP on-prem gap-closure plan by reducing retry-heavy RAP authoring failures and improving end-to-end behavior-pool workflow support.

## What changed
- Added deterministic RAP preflight validation for TABL/BDEF/DDLX/DDLS in `SAPWrite` with a `preflightBeforeWrite` toggle.
- Added `SAPWrite` action `scaffold_rap_handlers` with dry-run and auto-apply support.
- Improved behavior-pool handling by scanning/patching class includes (`main`, `definitions`, `implementations`) and fixed deferred class parsing in handler extraction.
- Added stronger write/activation hints (object-exists update hint, behavior-pool `[?/011]` remediation, per-object batch activation status).
- Updated RAP-related skills and docs (`docs_page/tools.md`, roadmap, CLAUDE references).
- Moved the implementation plan to completed: `docs/plans/completed/rap-onprem-agent-gap-closure.md`.

## Verification
- `npm test` ✅
- `npm run typecheck` ✅
- `npm run lint` ✅
- `npm run test:integration -- --run tests/integration/adt.integration.test.ts -t "RAP"` requires configured `TEST_SAP_*` / `SAP_*` credentials in this shell; status is documented in the completed plan notes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -143,6 +143,8 @@ src/
 │   ├── gcts.ts                 # gCTS Git backend client (/sap/bc/cts_abapvcs/*, JSON)
 │   ├── abapgit.ts              # abapGit ADT bridge client (/sap/bc/adt/abapgit/*, XML/HATEOAS)
 │   ├── cds-impact.ts           # CDS downstream impact classifier (RAP-oriented buckets)
+│   ├── rap-preflight.ts        # Deterministic RAP static-rule validator (TABL/BDEF/DDLX/DDLS)
+│   ├── rap-handlers.ts         # RAP handler signature extraction/matching/injection helpers
 │   ├── ui5-repository.ts       # UI5 ABAP Repository OData client
 │   ├── flp.ts                  # FLP PAGE_BUILDER_CUST OData client
 │   └── transport.ts            # CTS transport management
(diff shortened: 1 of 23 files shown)


marianfoo:
Reviewed the implementation in detail and verified it against the live test system (S/4HANA 2023 on-prem, A4H). Two bugs found and fixed in ae653b1:

### Bug 1: Factory/static action variants silently skipped

The action regex `^\s*(?:internal\s+)?action(?:\s*\([^)]*\))?\s+...` only matched `action` and `internal action`. Real BDEFs also use:

- `factory action copyTravel [1];` — factory action
- `static action foo;` — static action
- `static factory action bar;` — static factory

These were silently dropped from the requirements list, so `scaffold_rap_handlers` would claim "applied" without actually handling them. Verified against `/DMO/I_TRAVEL_M` (managed travel demo BDEF) which has both `factory action copyTravel [1];` and `internal action ReCalcTotalPrice;`.

### Bug 2: `RESULT result` was always emitted

The generator unconditionally appended `RESULT result` to every action signature. BDEF actions without a `result [...]` clause (common for factory/internal actions) would get an incorrect signature that doesn't match what ABAP expects.

Compared generated signatures against SAP's real handler class `/DMO/BP_TRAVEL_M/includes/implementations`:

| BDEF line | Real handler | Before fix | After fix |
|-----------|--------------|------------|-----------|
| `action ( features: instance ) acceptTravel result [1] $self;` | `METHODS ... RESULT result.` | `... RESULT result.` ✓ | `... RESULT result.` ✓ |
| `internal action ReCalcTotalPrice;` | `METHODS ... .` (no RESULT) | `... RESULT result.` ✗ | `... .` ✓ |
| `factory action copyTravel [1];` | `METHODS ... .` (no RESULT) | (not detected) ✗ | `... .` ✓ |

Also handles multi-line action declarations like `action foo\n  result [1] $self;` by scanning forward to the `;` terminator.

### Verified and kept as-is

- **Include fallthrough (main → definitions → implementations)**: confirmed correct against real behavior pools `ZBP_R_FBCLUBTP` and `/DMO/BP_TRAVEL_M`. Handler classes live in `implementations` (CCIMP); `main` holds the abstract public class; `definitions` is usually the empty template. The filter-skip-fallthrough pattern in `scaffold_rap_handlers` handles all three correctly.
- **`lhc_{alias}` naming**: matches both customer (`ZBP_R_FBCLUBTP` → expects `lhc_fbclubtp`) and SAP standard (`/DMO/BP_TRAVEL_M` → uses `lhc_travel`) conventions. When the class doesn't exist, the requirement is surfaced in `skipped` with a clear message so the user can correct manually.

All 2135 unit tests pass (2 new tests added for the action variants).

PR #179, opened 2026-04-22
fix: SAPActivate phantom success + CLI/server alignment gaps (NW 7.50)

samibouge:
# SAPActivate phantom success + CLI/server alignment gaps (NW 7.50)

PR given for reference, feel free to re-prompt it.

## Summary

`SAPActivate` reported *"Successfully activated"* for an ABAP class that was provably inactive (still failing to compile in Eclipse, visible as inactive in SAP GUI). The false-positive masked five independent bugs, each of which independently should have exposed the problem. The combination produced a silent no-op.

Reproducer:

```bash
npx tsx src/cli.ts activate CLAS ZCL_XXX
# before fix:
#   → Successfully activated CLAS ZCL_XXX.
# after fix:
#   → Activation failed for CLAS ZCL_XXX.
#     Errors:
#     - Activation did not complete — ZCL_XXX is still inactive.
#     - Activation did not complete — ZCL_XXX …EXECUTE_ACTION is still inactive
#       (pending changes owned by <USERID>).
#     Server syntax check (inactive):
#       - Line 52: Result type of the functional method "LX_ERR->IF_MESSAGE~GET_TEXT"
#                  cannot be converted into the type of formal parameter "IV_MSG_TEXT".
```

### 1. `parseActivationResult` misreads `<ioc:inactiveObjects>` as success

**File:** [src/adt/devtools.ts](src/adt/devtools.ts) — `parseActivationResult()`

SAP's `/sap/bc/adt/activation?method=activate` endpoint can return three response shapes:

- Empty body — success
- `<chkl:messages>` with `<msg>` entries — errors/warnings
- `<ioc:inactiveObjects>` with `<ioc:entry>` — the object is still inactive (and SAP tells us which transport / user owns the pending edit)

The parser only inspected `<msg>` nodes. A response with zero `<msg>` elements but several `<ioc:inactiveObjects>` entries was read as *"no errors → success"*. vibing-steampunk's Go client correctly treats this shape as failure.

**Captured response body (NW 7.50):**

```xml
<ioc:inactiveObjects xmlns:ioc="http://www.sap.com/abapxml/inactiveCtsObjects">
  <ioc:entry><ioc:object/><ioc:transport ioc:user="<USERID>" …/></ioc:entry>
  <ioc:entry><ioc:object …><ioc:ref adtcore:uri="…/ZCL_XXX" …/></ioc:object>…</ioc:entry>
  <ioc:entry><ioc:object …><ioc:ref adtcore:uri="…EXECUTE_ACTION" …/></ioc:object>…</ioc:entry>
</ioc:inactiveObjects>
```

### 2a. `getInactiveObjects` hardcoded to the Cloud endpoint path

**File:** [src/adt/client.ts](src/adt/client.ts) — `getInactiveObjects()`

Newer SAP (S/4, BTP) exposes the list at `/sap/bc/adt/activation/inactive`; NW 7.50 uses `/sap/bc/adt/activation/inactiveobjects`. The client hardcoded the new path and returned 404 on older systems. The NW 7.50 discovery fixture ([tests/fixtures/xml/discovery-nw750.xml:1129](tests/fixtures/xml/discovery-nw750.xml#L1129)) and [scripts/probe-all.ts:297](scripts/probe-all.ts#L297) both had the correct path — the client method didn't consult either.

### 2b. `parseInactiveObjects` only understood the feed-wrapped shape

**File:** [src/adt/xml-parser.ts](src/adt/xml-parser.ts) — `parseInactiveObjects()`

The parser iterated `<entry>` nodes and expected `<objectReference>` beneath each — the feed shape returned by newer systems. NW 7.50 returns a flat list:

```xml
<adtcore:objectReferences>
  <adtcore:objectReference adtcore:uri="…" adtcore:type="CLAS/OC" adtcore:name="…"/>
  <adtcore:objectReference …/>
  …
</adtcore:objectReferences>
```

No `<entry>` nodes → parser returned `[]` even on a 200 with content. Combined with 2a, this made `SAPRead type=INACTIVE_OBJECTS` useless on the affected system.

### 3a. `SAPDiagnose syntax` had no `version` parameter

**Files:** [src/handlers/schemas.ts](src/handlers/schemas.ts) (SAPDiagnoseSchema), [src/handlers/intent.ts](src/handlers/intent.ts) (`handleSAPDiagnose`), [src/handlers/tools.ts](src/handlers/tools.ts)

The low-level `syntaxCheck()` in `devtools.ts` already supported `{ version: 'active' | 'inactive' }`, but the MCP tool schema didn't expose it and the handler always defaulted to `active`. SAP's `chkrun:version` attribute controls which version the compiler validates, so for any object with pending changes the caller could not reach the real compile error through the tool.

### 3b. `parseSyntaxCheckResult` missed NW 7.50's message shape

**File:** [src/adt/devtools.ts](src/adt/devtools.ts) — `parseSyntaxCheckResult()`

The parser looked for `<msg>` with `@_line`/`@_col` attributes. The NW 7.50 `/sap/bc/adt/checkruns` endpoint returns `<chkrun:checkMessage>` with position encoded inside the `uri` attribute as `uri="…/source/main#start=LINE,COL"`. Two gaps in one: wrong element name, and the line/column were in a fragment component of another attribute.

## Fixes

All changes verified end-to-end against the reproducer system.

| # | File | Change |
|---|---|---|
| 1 | [src/adt/devtools.ts](src/adt/devtools.ts) | `parseActivationResult` now detects `<ioc:inactiveObjects>` entries with an `<object>` child and maps them to error-severity details (includes owning user from `ioc:transport[@_user]`). |
| 1+ | [src/handlers/intent.ts](src/handlers/intent.ts) | On activation failure the handler calls a new `inactiveSyntaxDiagnostic()` and appends per-object compiler errors from the inactive version to the error message. Applied to both single and batch paths. |
| 2a | [src/adt/client.ts](src/adt/client.ts) | `getInactiveObjects` tries `/activation/inactive` first and falls back to `/activation/inactiveobjects` on 404. |
| 2b | [src/adt/xml-parser.ts](src/adt/xml-parser.ts) | `parseInactiveObjects` handles both feed-wrapped and flat `<adtcore:objectReferences>` shapes. |
| 3a | [src/handlers/schemas.ts](src/handlers/schemas.ts), [src/handlers/intent.ts](src/handlers/intent.ts), [src/handlers/tools.ts](src/handlers/tools.ts) | Added `version: 'active' \| 'inactive'` to `SAPDiagnose action=syntax`; plumbed through to `syntaxCheck({ version })`. |
| 3b | [src/adt/devtools.ts](src/adt/devtools.ts) | `parseSyntaxCheckResult` now also collects `<chkrun:checkMessage>` nodes and extracts line/column from `uri="…#start=LINE,COL"`. |

### Refactor done as part of #1+

`tryPostSaveSyntaxCheck()` was split into:

- `inactiveSyntaxDiagnostic(client, type, name)` — the reusable primitive (no type filter)
- `tryPostSaveSyntaxCheck(client, type, name)` — the DDIC-only gate, unchanged externally

## CLI / MCP server alignment

[src/cli.ts](src/cli.ts) exposes two generic entry points (`call`, `tools`) plus a small set of ergonomic shortcuts. The shortcuts fall behind the MCP tool schemas in several places — for example, the `syntax` shortcut does not expose the new `version` parameter, so a user cannot reach the inactive compiler error without dropping to `call SAPDiagnose --arg version=inactive`.

### Shortcut coverage matrix (12 MCP tools)

| MCP tool | CLI shortcut | Knobs exposed on shortcut | Knobs in tool schema but NOT on shortcut |
|---|---|---|---|
| `SAPRead` | `read <type> <name>`, `source <type> <name>` | `type`, `name`, `--flat` | all other `SAPReadSchema` fields (`INACTIVE_OBJECTS` type, `TABLE_CONTENTS`, `VERSIONS`, pagination, etc.) |
| `SAPSearch` | `search <query>` | `query`, `--max` | `objectType`, `packageName`, type scope filters |
| `SAPWrite` | — | — | all (create/update/delete/edit_method, package, source, transport) |
| `SAPActivate` | `activate <type> <name>` | `type`, `name` | `preaudit`, `version`, `objects` (batch), `publish_srvb`/`unpublish_srvb` actions, `service_type` |
| `SAPNavigate` | — | — | all (`definition`, `references`, `completion`, `hierarchy`) |
| `SAPQuery` | `sql <query>` | `query` | `maxRows`, row offsets, other actions |
| `SAPTransport` | — | — | all (`list`, `history`, `info`, `release`) |
| `SAPGit` | — | — | all |
| `SAPContext` | — | — | all |
| `SAPLint` | — (the `lint <file>` CLI command is a **local** offline linter, not the `SAPLint` tool) | — | all (lint/format on live objects, formatter settings) |
| `SAPDiagnose` | `syntax <type> <name>` | `type`, `name` (only the `syntax` action) | `version` ← just added, not wired to CLI; `unittest`, `atc`, `dumps`, `traces`, `quickfix`, `apply_quickfix`, `system_messages`, `gateway_errors` |
| `SAPManage` | — | — | all |

## Why did tests not catch this?

Four blind spots combined:

1. No test activates an object that should fail and asserts the failure surfaces.
2. E2E fixtures ([tests/e2e/fixtures.ts](tests/e2e/fixtures.ts)) are all valid `$TMP` objects with clean source.
3. Complex-object activation tests go through `activateBatch()` (RAP stacks), bypassing the single `activate()` path that had the bug.
4. Unit tests assert the parser's handling of mock responses but never assert the outgoing request body or the post-activation state.

A single E2E test that: creates a class with a deliberate type error → attempts `SAPActivate` → asserts failure → asserts the returned message names the user and includes the compiler text — would have caught all five bugs.

## Open follow-ups

- [ ] Consider renaming the local `lint` command to `lint-local` or `offline-lint` to free `lint` for `SAPLint` (the MCP tool), or collapse it under a `local` sub-group.
- [ ] Auto-generate at least the required/optional field hints from the Zod schemas so CLI help stays in sync when someone adds a new `SAPDiagnose action=x` or a new `SAPRead type=y`.
- [ ] Clean up XML entity residue in `SyntaxMessage.text` (fast-xml-parser leaves `&quot;` / `&gt;` in attribute values; either post-process in the parser or configure the parser to decode entities in attributes).
- [ ] Add unit tests for each of the five parser/handler fixes.
- [ ] Add an E2E regression test: activate an object with a deliberate inactive compile error, assert failure and the specific message surface.
- [ ] Decide direction on CLI/server alignment (see matrix above). This is a UX-level change best treated as its own follow-up.


samibouge:
Additionally, VSP already performs a pre-write syntax check via ADT checkruns and an optional auto-activation after each write.

cf. https://github.com/oisee/vibing-steampunk/blob/1989ce143851125bdd21154fab750f047f5c4151/pkg/adt/workflows_edit.go#L320

cf. https://github.com/oisee/vibing-steampunk/blob/1989ce143851125bdd21154fab750f047f5c4151/pkg/adt/workflows_edit.go#L397

I prototyped this but consider it risky: when the agent plans to change multiple interdependent files, the syntax check on the first write will inevitably fail. A safer approach is to let the agent signal when it is done editing and explicitly request activation. Syntax errors are then surfaced at that point (which my PR now enables).

I considered just giving the information as a warning instead of a blocking error. From my experience Claude is good at ignoring warnings that are not relevant, but then it would waste tokens and dilute error signals.

samibouge:
## NW 7.50 compatibility note

All the activation and lock-conflict fixes in this PR were developed and tested exclusively against NW 7.50. The ADT framework on NW 7.50 has significant gaps in error reporting:

- **Lock conflicts surface as auth errors**: `CL_SEU_ADT_RES_ACTIVATION` raises `CX_ADT_RES_NO_ACCESS` (→ HTTP 403) for lock conflicts instead of `CX_ADT_RES_CONFLICT` (→ 409). With cookie auth, ICM further transforms this to 401/400 with a generic HTML login page — discarding the lock owner, object name, and all structured info that the ABAP layer had available.

- **Activation preaudit is implicit**: The two-phase handshake (`preauditRequested=true` → inactive objects prompt → `preauditRequested=false`) may behave differently or be unnecessary on newer releases where activation handles sub-includes transparently.

- **Status codes are inconsistent**: The same lock conflict returns 400, 401, or 403 depending on auth method and retry state. Our detection relies on matching `"Logon Error Message"` in the HTML body — a heuristic that works on NW 7.50 but may not apply on S/4 or BTP ABAP where SAP likely fixed the exception class and returns a proper structured 409/423 response.

On more recent releases (S/4HANA, BTP ABAP Environment), SAP has likely matured the ADT error handling — using the correct exception classes, returning structured XML error bodies with lock owner info, and mapping lock conflicts to proper HTTP status codes. **The NW 7.50 workarounds (HTML body sniffing, preaudit handshake, 400/401/403 reclassification) may need to be gated behind the detected system release** to avoid interfering with the improved error handling on newer systems. ARC-1 already has release detection infrastructure via `src/adt/features.ts` (`detectSystemCapabilities`) that could be used to conditionally enable these workarounds only for older releases.

marianfoo:
Thank you! I saw some more commits and comments in here. Just let me know when you´re ready so i can review it

samibouge:
The activation part on NW 7.50 is now fine. Could you test it on S/4?

Modern NW releases handle the locking error messaging better and it makes the code a bit difficult to have always fallbacks, so I gated the "weird lock handling" to release 7.50

On my tests on NW 7.50 I was able to successfully test
- if the object is locked it gets detected automatically (but on NW 7.50 I don't get the lock owner's name)
- activation and mass activation works
- syntax errors during activation and mass activations are surfaced

I've replaced VSP with ARC-1 as my daily MCP so I had to make small other fixes as well that are not yet in this PR.
These fixes are more "breaking" so I'd like to test on S/4 before submitting another PR.

marianfoo:
@samibouge Sure, we go step by step, let me review this first and then we can go through the other future PRs

samibouge:
FYI all my tests were done with cookie auth, other auth methods might have edge cases for the lock handling as well.

marianfoo:
will review this next, had to do a authorization refactor to make it more simple 
might want to remove the readme update from this pr

samibouge:
My mistake, I forgot I did the PR from the main branch (smart).
I've reverted the README change

marianfoo:
please test again with my changes

samibouge:
Thanks, I've done some manual testing against NW 7.50 but the error message has a few problems - I've aligned now the syntax and added the ADT url path to make it clear as CLAS objects have several sections.

Before:
```
Activation failed for CLAS ZCL_DEMO.

Errors:
- [line 1] Method "GET_DATA2" does not exist. There is, however, a method with the similar name "GET_DATA".
Server syntax check (inactive):
  - Line 34: Method &quot;GET_DATA2&quot; does not exist. There is, however, a method with the similar name &quot;GET_DATA&quot;.
```

After:
```
Activation failed for CLAS ZCL_DEMO.

Errors:
- [line 34] Method "GET_DATA2" does not exist. There is, however, a method with the similar name "GET_DATA". (/sap/bc/adt/oo/classes/zcl_demo/source/main#start=34,9)
Server syntax check (inactive):
- [line 34] Method "GET_DATA2" does not exist. There is, however, a method with the similar name "GET_DATA". (/sap/bc/adt/oo/classes/zcl_demo/source/main#start=34,9)
```

I'm also deduplicating the message but only on single activation because it was easier to detect if the error after activation was filled or not.

I've update the e2e test activation-failure.e2e.test.ts with batch activate and a check that the line number is right.
This e2e test was validated against both NW 7.50 and S/4HANA 2022.

PR #182, opened 2026-04-23
fix: detect HANA via S4CORE/HDB components

frozenraindrop:
## Problem

SAPManage probe reported hana.available: false on a confirmed S/4HANA system. The hanainfo endpoint was absent (404), and the component-based fallback that was meant to catch this returned an empty list due to a missing Accept header — so both signals silently failed.

## Root cause 1 — hanainfo 404 on some S/4HANA releases

probeFeatures() probes HANA via GET /sap/bc/adt/ddic/sysinfo/hanainfo. On some S/4HANA releases this ICF node is not activated. A 404 is treated as "no HANA", producing a false negative even on systems where HANA is the only supported database.

## Root cause 2 — missing Accept header on /sap/bc/adt/system/components

Both detectSystemFromComponents() (features.ts) and getInstalledComponents() (client.ts) fetched /sap/bc/adt/system/components without an Accept header. On the test system, omitting the header caused SAP to return a response the XML parser could not find a <feed> root in, silently yielding []. The required header is Accept: application/atom+xml;type=feed.

### 1 — S4CORE/HDB component fallback (src/adt/features.ts)

New exported helper detectHanaFromComponents() inspects the already-fetched installed-components list (zero extra HTTP requests — the call is in-flight for system-type detection anyway):

- Component name matching /HDB|HANA/i -> hasHana: true, hanaRelease from that component's release field.
- Component name S4CORE -> hasHana: true, hanaRelease undefined (S/4HANA always runs on SAP HANA; DB release cannot be inferred from the software-component layer).

When the endpoint probe returns false but components confirm HANA, the result is overridden to true (auto mode only — forced on/off configs are unaffected).

### 2 — Content negotiation (src/adt/client.ts + src/adt/features.ts)

Both call sites now send Accept: application/atom+xml;type=feed. A 406 guard in getInstalledComponents() returns [] instead of throwing, so systems using a different MIME scheme degrade gracefully.

## Known limitation

On the specific test S/4HANA 2023 system that triggered this investigation, even with the correct Accept header the endpoint returns HTTP 200 with an empty Atom feed (no <entry> elements, no SU53 auth denials). The S4CORE fallback therefore does not trigger on that system and hana.available remains false.

The reference ADT client library (github.com/marcellourbani/abap-adt-api) does not implement /sap/bc/adt/system/components at all, suggesting there is no universally reliable ADT signal for HANA presence across all S/4HANA releases. The v2.5.0 "core discovery for hana 1909" fix in that library addressed empty workspace arrays in a different discovery flow, not HANA detection.

The detection logic (S4CORE => HANA, HDB component => HANA) is correct by design, but whether it fires in practice depends on the components feed being populated — something we cannot verify on this system. Input from maintainers or contributors with access to systems where the feed is non-empty would be valuable.

## Tests (tests/unit/adt/features.test.ts)

Nine new unit tests:
- detectHanaFromComponents: HDB direct, HANA in name, S4CORE alone, no indicators, empty list, case-insensitive match, HDB priority over S4CORE
- probeFeatures: S4CORE in components overrides a 404 hanainfo (available: true); no HANA indicators + 404 hanainfo stays available: false

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -223,7 +223,7 @@ tests/
 | Add / modify auth combination rule | `src/server/config.ts` (validateConfig at ~line 305), `src/server/types.ts` (ServerConfig), `tests/unit/server/config.test.ts`, `docs/enterprise-auth.md` (Coexistence Matrix) |
 | Add Layer B auth mechanism | `src/adt/http.ts` (applyAuthHeader at ~line 830, fetchCsrfToken at ~line 669), `src/server/server.ts` (buildAdtConfig — perUser flag), `tests/unit/adt/http.test.ts` |
 | Add safety config option | `src/adt/safety.ts`, `src/server/config.ts`, `src/server/types.ts` |
-| Add feature probe | `src/adt/features.ts` |
+| Add feature probe | `src/adt/features.ts` (`PROBES` array — single endpoint per feature; HTTP-status classification handled centrally by `classifyFeatureProbeStatus`: 2xx/400/405/5xx → available, 401/403/404 → unavailable with diagnostic reason. Surface the reason via `FeatureStatus.message`.) |
 | Add feature-gated write guard | `src/handlers/intent.ts` (checkRapAvailable pattern), `src/adt/features.ts` |
 | Add E2E test | `tests/e2e/`, helpers in `tests/e2e/helpers.ts`, fixtures in `tests/e2e/fixtures.ts` |
 | Add/modify E2E fixture | `tests/e2e/fixtures.ts` (define object), `tests/fixtures/abap/` (source file), `tests/e2e/setup.ts` (sync logic) |
--- a/src/adt/client.ts
+++ b/src/adt/client.ts
@@ -18,7 +18,7 @@
 
 import type { AdtClientConfig } from './config.js';
 import { defaultAdtClientConfig } from './config.js';
-import { isNotFoundError } from './errors.js';
+import { AdtApiError, isNotFoundError } from './errors.js';
 import { AdtHttpClient, type AdtHttpConfig } from './http.js';
 import { checkOperation, OperationType, type SafetyConfig } from './safety.js';
 import { Semaphore } from './semaphore.js';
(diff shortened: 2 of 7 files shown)


marianfoo:
@frozenraindrop i made some changes so it worked on my test system, can you test that on your side again please, if its ok on your site i will merge

frozenraindrop:
This PR closed the first two gaps. But in my scenario I still got HANA false, since I'm connecting to a system where the components feed returns an empty feed even with the correct \`Accept\` header. And \`hanainfo\` returns 404. 
 
 We found that the ADT discovery document (\`/sap/bc/adt/discovery\`) is a reliable last resort: NHI (Native HANA
  Integration) workspaces register themselves under \`/sap/bc/adt/nhi/*\` in the discovery XML, and those entries only
make sense on HANA-backed systems. Since the discovery document is already fetched at startup for MIME negotiation, this
  adds zero extra HTTP requests.

  Confirmed working on the system in question. Follow-up fix: https://github.com/marianfoo/arc-1/pull/195

marianfoo:
currently thinking about what is this actually for, do you use and need this hana indicator? i lean more towards remove this hana discovery

frozenraindrop:
I am not currently using it myself explicitly. However, I believe the LLM should be aware of it.
For example, in [vscode_abap_remote_fs](https://github.com/marcellourbani/vscode_abap_remote_fs) the relevant HANA skills defined, such as https://github.com/marcellourbani/vscode_abap_remote_fs/blob/master/client/media/skills/abap-performance-hana/SKILL.md instruct the LLM to use one or the other depending on the HANA discovery:

> description: ABAP performance best practices for S/4HANA and HANA database systems.Use when writing or reviewing ABAP code on HANA-based systems.IMPORTANT: First use the SAP system info tool to check the system type — if the system is ECC or runs on a traditional database (Oracle, DB2, MSSQL), load the abap-performance-ecc skill instead. Covers code pushdown, CDS views, AMDP, advanced SQL, and HANA-optimized patterns.

marianfoo:
ok, lets merge it as is, might be useful later

PR #187, opened 2026-04-26
docs: extension-points case study

marianfoo:
## Summary

- Adds §14 to [docs/research/tool-extension-points.md](https://github.com/marianfoo/arc-1/blob/claude/extension-nw750-case-study/docs/research/tool-extension-points.md) — a worked case study evaluating the FEAT-61 extension concept against the [samibouge:arc-1:feat/nw750-version-fix](https://github.com/marianfoo/arc-1/compare/main...samibouge:arc-1:feat/nw750-version-fix) branch (18 commits).
- Updates the [FEAT-61 roadmap entry](https://github.com/marianfoo/arc-1/blob/claude/extension-nw750-case-study/docs_page/roadmap.md#feat-61) to label samibouge's fork as the motivating example and link to §14.
- **No code change.**

## Key conclusions

- **17 of 18 commits are vanilla upstream candidates** (several already merged via [#179](https://github.com/marianfoo/arc-1/pull/179)). Bug fixes, safety guards, NW 7.50 URL routing on the standard ADT API, opt-in pre-write check, cookie reload, CLI subcommand. None need an extension model.
- **Exactly one commit** ([\`8dedcb0\`](https://github.com/marianfoo/arc-1/commit/8dedcb0): NW 7.50 dump detail via custom ICF endpoint) is the case the user pushed back on. It requires installing \`ZCL_ARC1_DUMP_HANDLER\` on the SAP system. Five concrete reasons not to merge this upstream are laid out in §14.2: permanent ABAP-class maintenance, deployment story breaks, worse failure modes, slippery-slope precedent, centralised-gateway pitch erodes.
- **The extension concept handles this cleanly.** The customer (samibouge) keeps the ABAP class in their own repo, ships a small TS plugin that wraps \`client.http.get('/sap/rest/arc1/dumps')\`, and admins opt in via \`ARC1_PLUGINS=...\`. ARC-1 core stays clean of \`useCustomDumpEndpoint(abapRelease)\` and the \`/sap/rest/arc1/dumps\` URL. A worked plugin sketch is in §14.3.

## Honest tension surfaced

The Phase 1+2 model in §6 lets plugins **add** \`Custom_*\` tools but not **replace** built-ins. Under that constraint, samibouge's seamless \"\`SAPDiagnose(action=\"dumps\")\` transparently uses the custom endpoint on NW 7.50\" UX is not directly reproducible — the LLM has to call \`Custom_GetNw750Dump\` instead. §14.4 lays out three answers (live with it / Phase 1.5 route hook / Phase 3 manifest routes) and recommends \"live with it for v1 and revisit only when a second real plugin author hits the same wall.\"

## Suggested next step (no commitment)

Reply to samibouge along the lines of §14.6: 13 of 18 commits are clean upstream PR candidates (some already merged); the dump-handler commit is the canonical \"wait for FEAT-61\" example.

## Test plan

- [ ] Spot-check markdown renders cleanly on the docs site (mkdocs)
- [ ] Verify all internal links in §14 resolve (commit shas, anchors, FEAT-29g, FEAT-61, research doc cross-references)
- [ ] Confirm the FEAT-61 roadmap entry's case-study link points to §14 anchor
- [ ] No source files modified — no \`npm test\` / \`npm run typecheck\` impact expected

### DIFF
--- a/docs/research/tool-extension-points.md
+++ b/docs/research/tool-extension-points.md
@@ -660,3 +660,259 @@ A focused reply along these lines (no commitment yet, since FEAT-61 is P3):
 > Thanks for the deep NW 7.50 work. About 13 of the 18 commits are clean upstream candidates and we'd be happy to take them as small focused PRs (some already merged via #179). The custom ICF endpoint commit (`8dedcb0`) we're not going to merge upstream — installing `ZCL_ARC1_DUMP_HANDLER` on every customer's SAP system is the line ARC-1 doesn't cross. We've sketched a plugin model in [docs/research/tool-extension-points.md](./tool-extension-points.md) (FEAT-61) that is purpose-built for exactly this: you keep the ABAP class in your own repo, ship a small TS plugin that wraps `client.http.get('/sap/rest/arc1/dumps')`, and admins opt in via `ARC1_PLUGINS=...`. Until FEAT-61 lands you can already run a private fork; once it lands you can publish the plugin separately without forking ARC-1.
 
 That answers the user-stated concern ("don't want custom code of mine") while also giving a concrete path that doesn't waste samibouge's work.
+---
+
+## 15. Case study: integrating dassian-adt v2.0 as an ARC-1 extension
+
+A concrete worked example of how each extension pattern from §5 would (or would not) accommodate a real third-party MCP server: [dassian-adt](https://github.com/messianic-swop450/dassian-adt) (the upstream repository `DassianInc/dassian-adt` no longer exists; this is the surviving fork from `messianic-swop450`, captured 2026-04-26).
+
+### 15.1 What dassian-adt is
+
+dassian-adt v2.0 is a competing TypeScript MCP server for SAP ABAP, MIT-licensed, ~270 KB of TypeScript. It is built on top of [`abap-adt-api`](https://github.com/marcellourbani/abap-adt-api) (Marcello Urbani's ADT HTTP client) and ships with ~39 tools across 12 domain handlers.
+
+| Property | Value |
+|----------|-------|
+| Stack | Node 18+, TypeScript 5.7, `@modelcontextprotocol/sdk` 1.28, `abap-adt-api` 7.1 |
+| License | MIT |
+| Tools | ~39 across 12 handler classes |
+| Architecture | One `BaseHandler` per domain; `index.ts` wires them; `validateAndHandle` dispatches |
+| Auth | Basic, OAuth (XSUAA, Entra ID), or built-in OAuth login HTML form |
+| Multi-system | Yes — `SAP_SYSTEMS` env var, `sap_system_id` auto-injected into every tool schema |
+| Transports | stdio + Streamable HTTP |
+| MCP capabilities | `tools`, `prompts`, `logging` (sampling and elicitation used internally) |
+
+### 15.2 Tool inventory (39 tools, 12 handlers)
+
+| Handler | Tools |
+|---------|-------|
+| `SourceHandlers` | `abap_get_source`, `abap_set_source`, `abap_edit_method`, `abap_set_class_include`, `abap_pretty_print`, `abap_revisions`, `abap_get_function_group` |
+| `ObjectHandlers` | `abap_create`, `abap_delete`, `abap_activate`, `abap_search`, `abap_activate_batch`, `abap_object_info` |
+| `DataHandlers` | `abap_query`, `abap_table` |
+| `DdicHandlers` | `ddic_element`, `ddic_references` |
+| `GitHandlers` | `git_repos`, `git_pull` |
+| `QualityHandlers` | `abap_syntax_check`, `abap_atc_run`, `abap_atc_variants`, `abap_where_used`, `abap_find_definition`, `abap_fix_proposals` |
+| `RapHandlers` | `rap_binding_details`, `rap_publish_binding` |
+| `RunHandlers` | `abap_unlock`, `abap_run` |
+| `SystemHandlers` | `login`, `healthcheck`, `abap_get_dump`, `abap_inactive_objects`, `abap_annotation_defs`, `raw_http` |
+| `TestHandlers` | `abap_create_test_include`, `abap_unit_test` |
+| `TraceHandlers` | `traces_list`, `traces_set_parameters`, `traces_create_config`, `traces_hit_list`, `traces_statements`, `traces_db_access`, `traces_delete`, `traces_delete_config` |
+| `TransportHandlers` | `transport_create`, `transport_assign`, `transport_release`, `transport_list`, `transport_info`, `transport_delete`, `transport_set_owner`, `transport_add_user`, `transport_contents` |
+
+### 15.3 Coverage vs ARC-1's 12 intent tools
+
+Most dassian-adt tools have a one-to-one ARC-1 equivalent — they just live as `action`/`type` parameters on a shared intent tool rather than as separate tools. The interesting subset is what is *not* yet in ARC-1:
+
+| dassian-adt feature | ARC-1 status | Notes |
+|---------------------|--------------|-------|
+| `raw_http` escape-hatch tool | None — by design | Conflicts with ARC-1's "every endpoint has a safety guard" invariant. dassian-adt's own description warns "NEVER use raw_http to POST to lock endpoints" — exactly the kind of foot-gun ARC-1 wants to make impossible. |
+| `abap_unlock` (manual unlock) | None | Useful for orphaned-lock recovery. Could be added as `SAPManage(action="unlock_object")`. |
+| `abap_create_test_include` | None | Convenience for testclass scaffolding. Sits naturally in `SAPWrite`. |
+| `traces_create_config` / `traces_set_parameters` / `traces_delete_config` | Partial — `SAPDiagnose(action="traces")` reads only | Trace lifecycle (create config → run → analyze → delete) is more granular in dassian-adt. |
+| `abap_atc_variants` (list available variants) | None | ARC-1 takes a single variant param. Listing available variants is a small read tool that fits `SAPDiagnose`. |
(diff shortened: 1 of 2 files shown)


samibouge:
I've checked the proposal, I believe option "E. Manifest-only" could fit already many use-cases.
Here's a potential plan with CRUD (but I think Read to begin with is enough) that allow customers to call their own Z* endpoint via HTTP handlers.

I have made tools in past with a BC-focus that could read SLG1, SM37 jobs etc... all of this should be feasible with the manifest option and enable agents to do a lot more troubleshooting. Based on your vision exposed in https://blog.zeis.de/posts/2026-02-16-librechat-enterprise-gpt/ end-users could ask some technical questions relative to an SAP process and the agent should be able to check idocs, jobs etc.. to answer.

On a side note, you should not allow any plugin to override standard tools otherwise it will complicate support. If needed, standard tools could be hidden/disabled via a new parameter, so that the LLM would pick the plugin tool.

# Declarative Plugin System for ARC-1 (Manifest-Only, No JS)

## Context

**The concrete scenario:**
1. `GET /sap/rest/abap_dumps?begin_date=...&end_date=...&user=...` → JSON array of dumps
2. `GET /sap/rest/abap_dumps/<dump_id>` → JSON dump detail

**The questions this design answers:**
- Schema formalization → JSON manifest format (section 2)
- LLM understanding of cross-action data flow → natural-language descriptions (section 3)
- LLM calling convention / namespace → `Custom_*` prefix, first-class MCP tool (section 4)
- Rights specification → per-action `policy` block (section 5)

---

## 1. Design Decision: Multi-Tool Manifests, Multi-Action Tools

Each manifest can define **multiple tools**, and each tool can have **multiple actions**. This mirrors the existing ARC-1 pattern (SAPDiagnose has dumps/traces/syntax/unittest/atc, SAPManage has features/probe/create_package, etc.) while letting a team bundle related tools in one file.

Rationale:
- Keeps tool count low (the entire point of ARC-1's 12-tool design)
- Cross-action data flow is natural — the LLM sees both actions in the same description
- `denyActions` works at action level: `Custom_AbapDumps.get_dump`
- Multiple tools per file is convenient for teams with many small endpoints

---

## 2. Manifest Format

`my-plugins.plugin.json` loaded via `ARC1_MANIFEST_PLUGINS=/path/to/my-plugins.plugin.json`

```jsonc
{
  "$schema": "https://arc1.dev/schemas/plugin-manifest-v1.json",
  "apiVersion": 1,
  "name": "my-team-tools",
  "version": "1.0.0",
  "description": "Custom diagnostic tools for our SAP landscape",

  "tools": [
    {
      "name": "Custom_AbapDumps",
      "description": "List and inspect ABAP short dumps from a custom REST endpoint.\n\nUse 'list_dumps' first to find dumps by date/user, then 'get_dump' with a dump_id from the results to see the full detail.",

      "actions": {
        "list_dumps": {
          "description": "List ABAP short dumps filtered by date range and user.",
          "request": {
            "method": "GET",
            "path": "/sap/rest/abap_dumps",
            "query": {
              "begin_date": "{{begin_date}}",
              "end_date": "{{end_date}}",
              "user": "{{user}}"
            },
            "headers": { "Accept": "application/json" }
          },
          "input": {
            "type": "object",
            "properties": {
              "begin_date": { "type": "string", "description": "Start date (YYYY-MM-DD)" },
              "end_date": { "type": "string", "description": "End date (YYYY-MM-DD)" },
              "user": { "type": "string", "description": "SAP username to filter (uppercase)" }
            }
          },
          "outputDescription": "JSON array of dumps. Each has: dump_id (pass to get_dump), timestamp, error_type, program, user.",
          "policy": { "scope": "read", "opType": "Read" }
        },

        "get_dump": {
          "description": "Get the full detail of a specific ABAP short dump.",
          "request": {
            "method": "GET",
            "path": "/sap/rest/abap_dumps/{{dump_id}}",
            "headers": { "Accept": "application/json" }
          },
          "input": {
            "type": "object",
            "properties": {
              "dump_id": {
                "type": "string",
                "description": "Dump ID from a list_dumps response"
              }
            },
            "required": ["dump_id"]
          },
          "outputDescription": "Full dump: error_type, program, include, line, what_happened, source_extract, call_stack, variables.",
          "policy": { "scope": "read", "opType": "Read" }
        }
      },

      "availableOn": "all",
      "hyperfocused": "custom_abap_dumps"
    }
    // More tools can follow in this array
  ]
}
```

---

## 3. Response Handling: Pass-Through by Default

Responses are **forwarded as-is** to the LLM. No parsing, no transformation.

The `AdtHttpClient.get()` returns `{ body: string, headers, status }`. The manifest interpreter takes `resp.body` and returns it directly as the MCP tool result text content. If the endpoint returns JSON, the LLM gets JSON. If it returns XML or plain text, the LLM gets that.

The `outputDescription` field on each action is appended to the action's description in the tool definition. It tells the LLM what shape to expect in the response — but ARC-1 does not parse or validate the actual response.

No `response.type` field needed. This is the simplest possible design and avoids unnecessary processing. If a future use case needs response transformation (field extraction, truncation), that's a Phase 2 TypeScript plugin concern.

---

## 4. Cross-Action Data Flow (The dump_id Problem)

Solved through **descriptions only** — no formal binding DSL:

1. **Tool-level description**: "Use 'list_dumps' first to find dumps, then 'get_dump' with a dump_id from the results"
2. **Per-param description**: dump_id says "from a list_dumps response"
3. **outputDescription on list_dumps**: "Each has: dump_id (pass to get_dump)"

This is the same approach all 12 built-in tools use. SAPDiagnose says "Without id: lists recent dumps. With id: returns details." LLMs handle this natively.

---

## 5. LLM Calling Convention

Plugin tools are **first-class MCP tools**, not wrapped in a meta-tool. The LLM sees:

```
Custom_AbapDumps(action="list_dumps", begin_date="2026-04-01", end_date="2026-04-28")
```

Then:
```
Custom_AbapDumps(action="get_dump", dump_id="20260428_153042_DEV_001")
```

**Naming rules (enforced at startup):**
- Plugin names MUST start with `Custom_` or `X_`
- `SAP*` is reserved for built-in tools
- Collision → fail-fast startup error

**Combined inputSchema exposed to LLM** (all action params merged, like SAPDiagnose):
```json
{
  "type": "object",
  "properties": {
    "action": { "type": "string", "enum": ["list_dumps", "get_dump"] },
    "begin_date": { "type": "string" },
    "end_date": { "type": "string" },
    "user": { "type": "string" },
    "dump_id": { "type": "string" }
  },
  "required": ["action"]
}
```

Token cost: ~240 tokens for a 2-action plugin. Compare: SAPDiagnose is ~600 tokens.

---

## 6. Rights / Authorization

Each action declares `policy: { scope, opType }`. At startup, these register as ACTION_POLICY entries:

```
Custom_AbapDumps           → { scope: "read", opType: "R" }
Custom_AbapDumps.list_dumps → { scope: "read", opType: "R" }
Custom_AbapDumps.get_dump   → { scope: "read", opType: "R" }
```

This plugs into the existing two-gate system:
1. **Server ceiling**: `checkOperation(safety, opType, actionName)` — blocks if the server config doesn't allow the op type
2. **User scope**: `hasRequiredScope(authInfo, scope)` — blocks if the user's token/profile doesn't include the scope

A write-capable plugin action would declare `{ scope: "write", opType: "Update" }` and require both `allowWrites=true` on the server AND `write` scope on the user.

**denyActions integration** (works out of the box):
```bash
SAP_DENY_ACTIONS=Custom_AbapDumps.get_dump    # block detail, allow listing
SAP_DENY_ACTIONS=Custom_AbapDumps             # block entire plugin
SAP_DENY_ACTIONS=Custom_*                     # block all plugins
```

**ListTools pruning**: `filterToolsByAuthScope()` already checks ACTION_POLICY per action — plugin actions are pruned identically to built-in ones.

---

## 7. URL Templating

- `{{param_name}}` in `path` → substituted from validated input, URL-encoded
- `{{param_name}}` in `query` values → substituted, URL-encoded; omitted if input param is undefined
- Path-template variables are **always required** (enforced at manifest load, even if input schema says optional)
- Query-template variables with undefined values are **silently dropped** (optional filters work naturally)

---

## 8. POST/PUT/DELETE Support (Full CRUD) -> lower priority from my point of view

For write endpoints, the manifest supports all HTTP methods:

```jsonc
{
  "create_dump_note": {
    "description": "Add an analyst note to a dump.",
    "request": {
      "method": "POST",
      "path": "/sap/rest/abap_dumps/{{dump_id}}/notes",
      "headers": {
        "Content-Type": "application/json",
        "Accept": "application/json"
      },
      "body": {
        "template": { "text": "{{note_text}}", "author": "{{author}}" }
      }
    },
    "input": {
      "type": "object",
      "properties": {
        "dump_id": { "type": "string", "description": "Target dump" },
        "note_text": { "type": "string", "description": "The note to add" },
        "author": { "type": "string", "description": "Author name" }
      },
      "required": ["dump_id", "note_text"]
    },
    "policy": { "scope": "write", "opType": "Update" }
  }
}
```

**Body templating**:
- `body.template`: A JSON object with `{{param}}` substitution. Serialized to JSON string for the request body.
- `body.raw`: A raw string template with `{{param}}` substitution. For non-JSON bodies (XML, plain text).
- No body → no request body (GET/DELETE without body).

**Safety for writes**: The framework enforces `checkOperation()` before dispatch. A POST action with `opType: "Update"` requires `allowWrites=true` on the server. CSRF token management is handled by `AdtHttpClient` automatically.

---

## 9. What the Manifest CANNOT Do (Intentionally)

- Multi-call orchestration (call A, use response to call B in one invocation)
- Response field extraction / JSONPath selectors / transformation
- Conditional logic in templates
- Custom auth headers beyond what AdtHttpClient provides
- Background work / cron / setInterval
- Open connections to non-SAP hosts

For any of these → use a TypeScript (Phase 2) plugin or in-tree contribution.

PR #195, opened 2026-04-27
fix:    add NHI discovery fallback for HANA detection

frozenraindrop:
## Problem

   Follows up on #182. That PR closed two gaps (missing `Accept` header + component-name broadening). This closes the
   third scenario acknowledged in the discussion: a system where the components feed returns an **empty feed** even
 with the correct header, and `hanainfo` returns 404. On that system, neither the endpoint probe nor the components
 fallback fires — HANA is reported as unavailable when it is in fact the only database.

   ## Solution

   The ADT discovery document (`/sap/bc/adt/discovery`) is a reliable third indicator. NHI (Native HANA Integration)
   workspaces register themselves under `/sap/bc/adt/nhi/*` in the discovery XML, and these entries are only present on
   HANA-backed systems.

   The discovery document is **already fetched at startup** for MIME negotiation — this adds zero extra HTTP requests.
   The fix threads an `nhiPresent` boolean through `DiscoveryFetchResult` and applies it as a third tier in
   `probeFeatures`, after the `hanainfo` probe and the components-based fallback.

   Confirmed on a real S/4HANA on-prem system where components were empty and hanainfo was 404 — the NHI workspace was
   present in the discovery XML and correctly identified HANA.

   ## Changes

   - `src/adt/discovery.ts` — `DiscoveryFetchResult` type, `hasNhiWorkspace()` (regex on raw XML, exported for
 testing),
   updated `fetchDiscoveryDocument` return type
   - `src/adt/features.ts` — third HANA detection tier in `probeFeatures`, `detectHanaFromDiscovery()` exported for
   testing
   - Tests covering NHI-present/absent, all-signals-absent (correctly unavailable), and
 401-on-hanainfo-with-NHI-present


marianfoo:
continue in #182

PR #196, opened 2026-04-28
feat: NW 7.50 compatibility fixes + cookie hot-reload

samibouge:
All these features have been developed based on my real-world ABAP development workflow. To be honest I'm surprised some of these problems existed from VSP. I think there is still a little bit of foundational work to clean up and harden the code against errors (imho your authz rework and extension framework go in the right direction).

1. I've removed the custom dump endpoint (I don't use it yet anyway and it would fit better your extension framework)

2. The release gate is based on a static release map but you mentioned previously you preferred probing. I would suggest starting with the static release map of features, anyway the filtering has been extracted into method filterByRelease so it can be changed easily later.

3. NW 7.50 has many critical bugs because the ADT API does insufficient checks (eg. if you give a task number instead of a transport number it still accepts it but fails halfway with no recovery possible, if you create a CDS view with a mixed case DDIC object name it accepts it etc...) - I've put guards against these in the code.

4. My code introduces the ability to read inactive versions, this might impact your work with caching/ETag which also mentions that feature.

Auto-generated summary of the squashed commits:
- SAPActivate phantom success detection + two-phase activation handshake
- Cookie hot-reload on stale auth (SAP_COOKIE_FILE re-read on 401)
- Release-gated dispatch for unavailable types/actions on older SAP_BASIS
- NW 7.50 TABL read routing, STRU write support, version param
- Mixed-case object name rejection on create
- MSAG create guard against task numbers on NW 7.50
- Activation error formatting (correct line numbers, decode entities, dedup)
- Opt-in pre-write SAP syntax check + inactive-version diagnostics
- 404 deletion-blocked error classification
- Lock-as-401 reclassification on createObject path
- CLI alignment: extract-cookies, cli-args, tool shortcuts

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -198,11 +198,11 @@ tests/
 | Add RAP deterministic preflight checks | `src/adt/rap-preflight.ts`, `src/handlers/intent.ts` (`runRapPreflightValidation`), `src/handlers/tools.ts`, `src/handlers/schemas.ts`, `tests/unit/adt/rap-preflight.test.ts` |
 | Add RAP behavior handler scaffolding logic | `src/adt/rap-handlers.ts`, `src/handlers/intent.ts` (`SAPWrite action=scaffold_rap_handlers`), `src/handlers/tools.ts`, `src/handlers/schemas.ts`, `tests/unit/adt/rap-handlers.test.ts` |
 | Add new tool type | `src/handlers/tools.ts`, `src/handlers/schemas.ts`, `src/handlers/intent.ts` |
-| Add/modify tool input schema | `src/handlers/schemas.ts`, `src/handlers/tools.ts` |
+| Add/modify tool input schema | `src/handlers/schemas.ts`, `src/handlers/tools.ts` — **three-file sync required** (see invariant below) |
 | Add DDIC domain/data element write | `src/adt/ddic-xml.ts`, `src/adt/crud.ts`, `src/handlers/intent.ts` |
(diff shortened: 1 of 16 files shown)


samibouge:
I'm not sure I agree with the decision to drop the ?version parameter.

I just had an example now where the LLM made a mistake while refactoring code, it dropped constants (as it often does). It checked on its own the active version to find back what it had dropped in the inactive version.

Of course this is anecdotal and there are other ways it could have done it, but the most efficient workflow I've found so far was for Claude Code to work directly on the ABAP server (with the risks it implies) so that it can handle together DDIC changes and code changes, and get direct feedback on mistakes.

In general if you look at ABAP development, the concept of inactive/active version is integral to the developer experience, and as such, should be exposed to the LLM as well.

marianfoo:
good point, let me think about this

marianfoo:
Hi @samibouge — closing this in favour of the split that landed across multiple PRs. Your contributions are credited as `Co-Authored-By` on each:

 - **#200** cookie hot-reload (merged 2026-05-08)
 - **#201** three-file sync + mixed-case object name rejection (merged 2026-05-08)
 - **#202** layered lock-conflict reclassification + MSAG transport-vs-task guard ([open, ready to merge])

 Decisions for the parts that took a different shape (STRU → collapsed into TABL by #219; SAPRead `version=active\|inactive` shipped via #186; static release gates replaced by the ARCH-01 plan) are recorded in the ADRs in #199.

 Thanks for the empirical NW 7.50 work — your `nw750-discovery-gap-analysis.md` is preserved in #199 as input for ARCH-01.

You can review now and create Issues/PRs if you find anything else

PR #205, opened 2026-05-04
ci: gate test workflow by PR title prefix

marianfoo:
## Summary
- Skip the test workflow on `docs:` and `chore:` PRs (saves SAP work processes — these PRs don't change runtime behavior).
- `chore:` PRs can still be tested on demand via **Actions → Test → Run workflow** (added `workflow_dispatch:` trigger).
- Auto-runs unchanged for `fix:`, `feat:`, `test:`, `refactor:`, `perf:`, etc., and for every push to `main`.

## How it works
A new `gate` job evaluates `github.event.pull_request.title` against the prefixes. The rest of the pipeline (`test`, `integration`, `e2e`, `reliability-summary`) chains off it via `needs:`, so a skipped gate cleanly skips the whole run instead of producing an empty "no results" reliability report.

Matches both `prefix:` and `prefix(scope):` forms (e.g. `chore(release):` is also gated).

## Caveats
- PR title changes after the initial `pull_request` event don't re-trigger the workflow on their own — the next push to the branch (or a manual dispatch) does.
- `reliability-summary` is now gated by `needs.gate.result == 'success'` instead of `always()`, so it won't try to summarize a skipped run. It still runs on test failures.

## Test plan
- [ ] Open this PR (`ci:` prefix) → workflow should still run normally.
- [ ] Verify a `docs:` PR (the docs PR opened just before this one) shows the gate job skipped and downstream jobs not running.
- [ ] On a chore-prefixed PR, confirm auto-run is skipped and **Run workflow** button works.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,8 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  # Manual trigger so chore: PRs (and any other skipped PR) can be tested on demand.
+  workflow_dispatch:
 
 # Only one SAP-hitting workflow at a time — prevents parallel runs from
 # different PRs/pushes from exhausting SAP work processes.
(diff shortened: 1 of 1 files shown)


marianfoo:
Folded into #161.

PR #225, opened 2026-05-08
fix!: use CreateCorrectionRequest endpoint for transport creation

sakirsek:
## Summary

`SAPTransport(action="create")` is unusable on NetWeaver 7.5x backends today: SAP rejects the legacy `POST /sap/bc/adt/cts/transportrequests` body with HTTP 400 `user action  is not supported`, and the existing `BACKEND_UNSUPPORTED` skip in the integration suite hides the gap rather than fixing it.

I dug into the SAP response (the message comes from `CL_ADT_TM_RESOURCE`, with empty `<properties/>` so it's not a T100-parameterised message) and confirmed by experiment that NW 7.5x ignores `tm:useraction` on the `POST .../transportrequests` body — that backend only honours `useraction` on PUT operations like `changeowner` / release.

The fix is to switch to the same path that Eclipse ADT, [`marcellourbani/abap-adt-api`](https://github.com/marcellourbani/abap-adt-api/blob/main/src/api/transports.ts), and [`jfilak/sapcli`](https://github.com/jfilak/sapcli/blob/master/sap/adt/cts.py) use: `POST /sap/bc/adt/cts/transports` with the `asx:abap` `CreateCorrectionRequest` schema (`DEVCLASS` / `REQUEST_TEXT` / `REF` / `OPERATION`). SAP infers `K`/`W`/`T` from the package's transport route in TADIR, so the legacy `transportType` argument becomes a no-op kept only for signature compatibility. The response body looks like `/com.sap.cts/object_record/<id>`; the transport ID is the last path segment.

Verified live against NW 7.51 SP02 (an on-prem dev system). Without the fix the `SAPTransport create` call fails 100% of the time; with it, transport creation now works end-to-end via the MCP tool.

## ⚠ Breaking change

`createTransport()` and `SAPTransport(action="create")` now require a target package. Without one the call returns an actionable error (`Package is required for "create" action — SAP needs DEVCLASS to determine the transport route.`) instead of a confusing 400 from SAP. The `package` parameter already existed on the tool schema for `check`; this PR documents it as required for `create` too.

## Test plan

- [x] `npx vitest run tests/unit/adt/transport.test.ts` — 68/68 pass with the rewritten body, response, content-type, and endpoint assertions
- [x] `npx vitest run tests/unit/handlers/intent.test.ts` — 458/458 pass, including a new "create errors when package is missing" case
- [x] `npx tsc --noEmit` — clean
- [x] Manual end-to-end: `SAPTransport(action="create", description="…", package="ZCAO")` now returns a fresh transport ID on NW 7.51 SP02 (verified with `SAPTransport(action="get")`)
- [ ] Re-run the integration suite against a system that exposes both Customizing and Workbench packages — I only had a Workbench-routed package available, so I dropped the per-type integration tests that relied on the legacy `tm:type` body assertion. They could come back later with `TEST_TRANSPORT_PACKAGE_W` / `TEST_TRANSPORT_PACKAGE_K` env vars.

### DIFF
--- a/src/adt/transport.ts
+++ b/src/adt/transport.ts
@@ -76,29 +76,67 @@ export async function getTransport(
   return match ?? null;
 }
 
-/** Create a new transport request */
+/**
+ * Create a new transport request via the ADT CreateCorrectionRequest endpoint.
+ *
+ * Uses POST `/sap/bc/adt/cts/transports` (the same endpoint Eclipse ADT and
+ * marcellourbani/abap-adt-api use). The legacy POST `/cts/transportrequests`
+ * path with a `<tm:root tm:useraction="newrequest">` body is rejected by
+ * NW 7.5x with HTTP 400 "user action  is not supported" because that release's
+ * `CL_ADT_TM_RESOURCE` only accepts `useraction` on PUT operations
+ * (changeowner, release). Verified against NW 7.51 SP02.
+ *
+ * SAP infers the transport type (K/W/T) from the package's transport route
+ * configured in TADIR, so the legacy `transportType` argument is ignored
+ * by the backend and kept here only for API compatibility.
+ *
+ * @param targetPackage required — DEVCLASS the transport is created for
+ * @param objectUrl optional — ADT object URL hint for transport-route lookup;
+ *                  the object is NOT locked or attached to the transport
+ */
 export async function createTransport(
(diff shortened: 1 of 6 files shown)


marianfoo:
closed in favor of #228, please check if this would help you

PR #267, opened 2026-05-11
feat: stable DCR signing key + 0/negative TTL = infinite

ClementRingot:
Title: feat: stable DCR signing key + 0/negative TTL = infinite

Body:

## Summary

Two related additions to the OAuth DCR layer to keep cached `client_id`s
valid across server-side events that would otherwise invalidate them:

1. **`ARC1_DCR_SIGNING_SECRET`** — new optional env var that decouples the
   HMAC signing key from the XSUAA `clientsecret`. Lets cached `client_id`s
   survive `cf deploy` operations that recreate the XSUAA service binding.
2. **`ARC1_OAUTH_DCR_TTL_SECONDS=0` now disables expiration entirely.**
   The 30-day default is unchanged; this only broadens the parser to treat
   non-positive values as "no TTL" — fully RFC 7591-compliant (see below).

## Background

While running ARC-1 on BTP CF behind XSUAA, we hit `invalid_client` errors
on MCP clients (GitHub Copilot CLI, Claude Desktop) even though the
stateless DCR design (#212) was supposed to make registrations survive
restarts. Investigation revealed two compounding causes — and crucially, a
**client-side amplifier** that turns either cause into a hard outage for
some MCP client implementations.

### Cause 1 — `cf deploy` rotates the XSUAA `clientsecret`

`cf deploy` of the MTA archive **recreates the XSUAA service binding** as
part of the deploy flow. Empirically observed via:

cf curl /v3/service_credential_bindings?service_instance_names=arc1-xsuaa
→ created: 2026-05-06T13:15:24Z   # right during the deploy

A new binding produces a new `clientsecret`. Because the DCR signing key
derives from that `clientsecret` (via HKDF-style HMAC), every redeploy
silently invalidates every previously-issued `client_id` — clients see
`oauth_client_lookup_failed` with `reason: "bad_signature"`.

### Cause 2 — TTL expiry of cached `client_id`s

With the upstream 30-day default, every cached `client_id` expires
exactly 30 days after issuance, even if the server never restarts and
the signing key never rotates. At that point, `getClient()` returns
`undefined` and the client sees `oauth_client_lookup_failed` with
`reason: "expired"` — same user-facing symptom as Cause 1, just
triggered by the server clock advancing past `iat + ttlSeconds`
instead of by a key rotation.

We hit this empirically on this branch: a `client_id` issued at
`iat = 2026-05-05 11:20:27 UTC` started failing on `2026-05-06 ~12:51 UTC`
(age ~25h, against the cherry-pick's 24h TTL — the upstream's 30d default
just delays the same failure to 30 days later).

### Cause 3 (amplifier) — Some MCP clients don't auto-re-register

RFC 7591 doesn't specify what a client should do on `invalid_client`.
Empirically observed on this branch:

| Client | Auto-re-registers on `invalid_client`? |
|---|---|
| Claude Desktop / Claude.ai | ✅ Yes |
| MCP Inspector | ✅ Yes |
| GitHub Copilot CLI | ❌ No (retries indefinitely with same dead `client_id`) |
| Copilot Studio | ❌ No (documented in upstream `xsuaa-setup.md`) |

For clients that auto-re-register, both causes above are invisible UX
glitches (1–2s extra round-trip). For clients that don't, both causes
become **hard outages**: the user gets stuck on `invalid_client` until
they manually clear the OAuth cache. This is the same symptom regardless
of the underlying reason (`bad_signature` vs `expired`) — the client
just keeps replaying the same cached `client_id`.

## Failure mode coverage

| Failure mode | Lookup `reason` | Fixed by |
|---|---|---|
| `cf deploy` (binding rotation) | `bad_signature` | `ARC1_DCR_SIGNING_SECRET` |
| Manual XSUAA rebind | `bad_signature` | `ARC1_DCR_SIGNING_SECRET` |
| 30-day TTL expiry | `expired` | `ARC1_OAUTH_DCR_TTL_SECONDS=0` |
| Explicit revocation by ops | `bad_signature` (intended) | n/a — that's the point |

The two changes together close every non-intentional invalidation path
for Copilot CLI / Copilot Studio users. Operators that need a
stricter compromise window keep the 30-day default (or set any positive
value clamped to `[60s, 90d]`).

### `ARC1_DCR_SIGNING_SECRET` (new)

- New optional env var / `--dcr-signing-secret` CLI flag
- When set: the HMAC signing key derives from this dedicated secret
  instead of the XSUAA `clientsecret`. Set once via `cf set-env`
  (which survives `cf deploy`) and registrations stay valid across
  redeploys. A startup `INFO` log confirms the dedicated secret is
  in use ("DCR signing key uses dedicated ARC1_DCR_SIGNING_SECRET").
- When unset: falls back to deriving the signing key from the XSUAA
  `clientsecret` (legacy behavior). No startup warning — the default
  is unchanged.
- Re-setting the value (with `cf restage`) is the explicit revocation
  knob — invalidates every outstanding `client_id` in one command.
- Recommended generation: `openssl rand -base64 48`.
- Startup INFO log gains a `dcrSigningSource: 'env' | 'xsuaa'` field
  for observability.

### `ARC1_OAUTH_DCR_TTL_SECONDS=0` ⇒ no expiration

- **Default is unchanged: 30 days.** Positive values still clamped to
  `[60s, 90d]`.
- **New semantic**: `0` (or any non-positive value) explicitly disables
  expiration. Previously, `0` was effectively unreachable — the parser
  only ran on truthy strings, so `"0"` was rejected as falsy; now the
  parser also handles `0` and negative values, mapping them to the
  "no TTL" case. When TTL is disabled, a startup `INFO` log confirms
  it ("DCR client_id TTL is disabled (ARC1_OAUTH_DCR_TTL_SECONDS=0)").
- Recommended setting when MCP clients in use don't auto-re-register
  on `invalid_client` (Copilot CLI, Copilot Studio,...) — a finite
  TTL in that mode just produces periodic 30-day outages without
  security gain, since revocation can be triggered explicitly via
  `ARC1_DCR_SIGNING_SECRET` rotation or `KDF_LABEL` bump.

### Configuration examples

**BTP CF (production)** — signing secret via `cf set-env` (survives
`cf deploy`), TTL via `mta.yaml` (not a secret):

```bash
# one-shot bootstrap of the signing secret
cf set-env arc1-mcp-server ARC1_DCR_SIGNING_SECRET "$(openssl rand -base64 48)"
cf restage arc1-mcp-server
```

```yaml
# mta.yaml — modules[arc1-mcp-server].properties
ARC1_OAUTH_DCR_TTL_SECONDS: "0"
```

**Local dev** (`.env` file):

```bash
ARC1_DCR_SIGNING_SECRET=<generated-with-openssl-rand-base64-48>
ARC1_OAUTH_DCR_TTL_SECONDS=0
```

**CLI flags** (equivalent):

```bash
arc1 --dcr-signing-secret "$SECRET" --oauth-dcr-ttl-seconds 0
```

### Why "no TTL" is RFC-compliant

RFC 7591 §3.2.1 (the Dynamic Client Registration response schema)
defines `client_secret_expires_at` and explicitly assigns the value
`0` the meaning **"will not expire"**:

> `client_secret_expires_at` — Time at which the client secret will
> expire **or 0 if it will not expire**.

So a registration that never expires is a first-class outcome in the
spec. Auth0, Keycloak, Okta, AWS Cognito, and Azure AD (Entra) all
default to non-expiring DCR registrations. The 30-day default kept by
this PR is a defense-in-depth choice, not a standard requirement; the
new `0` semantic just lets operators opt out of it RFC-cleanly.

## Security review

- `client_id` alone grants no access — the OAuth flow still requires
  user login at XSUAA + PKCE.
- `client_secret` is HMAC-derived deterministically from the
  `client_id`; leaking `client_id` ≈ leaking `client_secret`. Attack
  value of a stolen pair is bounded by the immutable `redirect_uris`
  in the signed payload (typically `http://localhost:<port>` for
  native MCP clients — only exploitable on the victim's own machine).
- HMAC-SHA256 truncated to 16 bytes (128 bits) — unchanged.
- `crypto.timingSafeEqual` for signature verification — unchanged.
- Revocation paths preserved: rotate `ARC1_DCR_SIGNING_SECRET`,
  rebind XSUAA, or bump `KDF_LABEL` (`arc1-dcr/v1` → `v2`).
- `ARC1_DCR_SIGNING_SECRET` constructor enforces non-empty; falsy
  values trigger explicit error at startup.

## Testing

- 2493 unit tests pass (4 new in `config.test.ts` for env-var parsing
  + precedence, plus updated TTL tests in `stateless-client-store.test.ts`
  covering default 30d, explicit `ttlSeconds=0`, and negative values).
- Typecheck + lint clean.
- Manually verified end-to-end on BTP CF: deployed with
  `ARC1_DCR_SIGNING_SECRET` set, observed `dcrSigningSource: env` in
  startup logs, then redeployed and confirmed Copilot CLI's cached
  `client_id` stayed valid (no `bad_signature` events).

## Rollout

For existing deployments wanting the new behavior:

```bash
# Stable signing key — survives cf deploy
SECRET=$(openssl rand -base64 48)
cf set-env arc1-mcp-server ARC1_DCR_SIGNING_SECRET "$SECRET"

# Optional: disable expiration if your client mix can't auto-re-register
cf set-env arc1-mcp-server ARC1_OAUTH_DCR_TTL_SECONDS 0
# (or declare it in mta.yaml properties — it's not a secret)

cf restage arc1-mcp-server

# Users clear their MCP client cache once for this rollout;
# subsequent cf deploys keep their client_ids valid.
```

For deployments unchanged: the 30-day TTL default is preserved, falling
back to the legacy XSUAA-clientsecret-derived signing key. No new startup
noise — both new INFO logs only fire when the corresponding opt-in is set.

Test plan

- Deploy to a non-prod BTP CF space without ARC1_DCR_SIGNING_SECRET set
- Verify startup log shows `dcrSigningSource: xsuaa` (and no DCR-related INFO line)
- Set ARC1_DCR_SIGNING_SECRET, restage; verify `dcrSigningSource: env` plus
  the "DCR signing key uses dedicated ARC1_DCR_SIGNING_SECRET" INFO line
- Connect with Copilot CLI / Claude Desktop, complete OAuth flow
- Run cf deploy again
- Verify the cached client_id is still accepted — no
  oauth_client_lookup_failed with bad_signature after redeploy
- Verify revocation: re-set ARC1_DCR_SIGNING_SECRET to a new value,
  cf restage, confirm previously-valid client_ids now fail with
  bad_signature
- Set ARC1_OAUTH_DCR_TTL_SECONDS=0, restage; verify the
  "DCR client_id TTL is disabled" INFO line appears, then verify a
  registration issued before the change stays valid past 30 days
  (clock-mocked unit test or live observation over a month)
- Conversely with a small TTL: verify Copilot CLI gets stuck on
  `invalid_client` after the TTL elapses (negative test).


marianfoo:
thank you! please review my additions

ClementRingot:
**Reviewed and approved**
Thanks for the thorough audit. Highlights:
- Empty/whitespace fallback with soft-warn — right call vs hostile startup crash on a typo'd `cf set-env`.
- 16-byte floor aligned with NIST SP 800-131A; `Buffer.byteLength` for UTF-8 correctness 
- `client_secret_expires_at` per RFC 7591 — missed it in my original PR.
- Orphan warning matches the existing `SAP_DISABLE_SAML` soft-warn pattern.
- +15 substantive tests, including the UTF-8 boundary edge case.

Merged locally, `npx vitest run tests/unit/server/` → 341 passing. Good to merge whenever you're ready.

PR #268, opened 2026-05-11
feat: warn on orphaned ARC1_DCR_SIGNING_SECRET + document DCR in MTA template

marianfoo:
## Summary

Two small follow-ups to [#267](https://github.com/marianfoo/arc-1/pull/267) (`feat: stable DCR signing key + 0/negative TTL = infinite`), addressing review concerns 2 and 5 raised against that PR.

**Note:** this branch is stacked on top of #267 — the diff currently shows both #267's commit and this one. Once #267 merges, the diff will collapse to the 6 files actually changed here.

### 1. `validateConfig` warning for orphaned `ARC1_DCR_SIGNING_SECRET`

In `src/server/config.ts:validateConfig`, after the existing `SAP_DISABLE_SAML` warning, add a parallel soft warning:

```
[warn] ARC1_DCR_SIGNING_SECRET is set but SAP_XSUAA_AUTH=false — the secret is unused. Unset it to reduce attack surface, or enable XSUAA OAuth proxy mode (SAP_XSUAA_AUTH=true).
```

**Why:** The signing secret is only consumed by the XSUAA OAuth proxy path ([src/server/http.ts:244](https://github.com/marianfoo/arc-1/blob/main/src/server/http.ts#L244) — `if (config.xsuaaAuth && xsuaaCredentials)`). Without XSUAA, the secret is silently parsed into `config.dcrSigningSecret` and then never read. Env vars routinely leak via `printenv`, `docker inspect`, crash dumps, and accidental log lines — a secret "set just in case" is dead-weight attack surface. The warning surfaces the misconfiguration so an operator can either unset it or enable XSUAA.

**Why warn (not throw):** A hard error would break legitimate staged-rollout patterns (set the secret in env, flip `SAP_XSUAA_AUTH=true` after verifying everything else loads), templated deployments where the same env spec is reused across modes, and CI matrices. The misconfiguration is operationally harmless — there's no security regression from an unread secret. Matches ARC-1's own precedent: throw on contradictions that silently break security (OIDC issuer/audience), warn on operationally-loose combinations (SAML on BTP).

Researched against RFC 9700 §2.5 (credential confidentiality), node-oidc-provider's convention (throw on inconsistent combinations, ignore stray unused config), and ARC-1's existing soft-warn pattern at `src/server/config.ts:587`.

### 2. `mta-overrides.mtaext.example` OAuth DCR section

New commented section between "Networking / public URL" and "CORS":

- `ARC1_OAUTH_DCR_TTL_SECONDS: \"0\"` commented example with rationale referencing the Copilot CLI / Cursor non-auto-re-register failure mode.
- Explicit `# NOTE:` block (no property line) explaining that `ARC1_DCR_SIGNING_SECRET` is intentionally **not** templated here — operators must use `cf set-env arc1-mcp-server ARC1_DCR_SIGNING_SECRET \"\$(openssl rand -base64 48)\"` instead, because MTA properties under `modules[*].properties` are rewritten on every `cf deploy`, which would rotate the signing secret each deploy and invalidate every cached `client_id` — defeating the whole purpose of the dedicated secret.

### 3. Doc cross-references

- `docs_page/enterprise-auth.md` "SAP Auth Coexistence Rules" gains rule 5 documenting the new warning.
- `docs_page/xsuaa-setup.md` "Stable DCR signing key" section gains a one-line note cross-referencing the warning.

## Testing

- 2832 unit tests pass (3 new in `config.test.ts`):
  1. Warns when `dcrSigningSecret` is set with `xsuaaAuth=false`.
  2. Does NOT warn when `dcrSigningSecret` is set with `xsuaaAuth=true`.
  3. Does NOT warn when `dcrSigningSecret` is unset.
- Typecheck + Biome lint clean.
- End-to-end manual check via `dist/`: warning fires under Case 1 and stays silent under Cases 2/3.
- YAML parser sanity check on the updated `mta-overrides.mtaext.example` (Python `yaml.safe_load`).

## Test plan

- [ ] Reviewer manually verifies the warning fires correctly:
  ```bash
  npm run build
  node -e "const {validateConfig} = require('./dist/server/config.js'); const {DEFAULT_CONFIG} = require('./dist/server/types.js'); validateConfig({...DEFAULT_CONFIG, dcrSigningSecret: 'x', xsuaaAuth: false});"
  # expect: [warn] ARC1_DCR_SIGNING_SECRET is set but SAP_XSUAA_AUTH=false ...
  ```
- [ ] Reviewer confirms `mta-overrides.mtaext.example` still parses as YAML.
- [ ] Reviewer reads the new mta-overrides NOTE for clarity (steering operators away from MTA properties is the whole point — verify the rationale lands).

## Rollout

No defaults change. No breaking change. Users who already configured #267 correctly (set `ARC1_DCR_SIGNING_SECRET` with `SAP_XSUAA_AUTH=true`) see no behavior change. Users who set the secret without XSUAA see one new stderr line at startup and can act on it.


marianfoo:
Superseded by additional commits on #267. The two changes here (validateConfig orphan warning + mta-overrides DCR section) are now part of commit bdf97808 on PR #267, alongside four additional hardening items surfaced by deeper audit:

- Robust ARC1_DCR_SIGNING_SECRET handling (empty/whitespace no longer crashes startup; dcrSigningSource label tracks effective secret)
- Soft warn on signing secret < 16 bytes (NIST SP 800-131A floor)
- RFC 7591 §3.2.1 client_secret_expires_at in /register response (0 = never expires)
- Doc fix: false 'logs a startup warning on legacy fallback' statement corrected

Full implementation plan: docs/plans/completed/pr267-hardening-bundle.md on PR #267.

Issue #313, opened 2026-05-29
Improve token usage on SAPRead


samibouge:
In my fork I improved the way SAPRead is handled based on how Claude likes to work with source code:

- by default SAPRead returns a compact representation:  only constants, types and method signatures (including line ranges)
- add a new parameter to read specific lines
- add a new parameter to allow GREPing the source code

Claude was trained already to work like this so in my fork it was able to use quite naturally these tools. It pairs well with #303 but #303 needs to be implemented first, otherwise the agent won't be able to act efficiently on it.

You may review the following commit and see if/how you would want to integrate the improvements
https://github.com/samibouge/arc-1/commit/91d71eafb0443ced7abcaace20b008aa502852c0

There are few subtleties in the way for example to handle multiple classes inside an include, so when grepping I return the full context (local class name + method name)

marianfoo:
implemented in #316

PR #307, opened 2026-05-28
feat: class-section surgery for CLAS

marianfoo:
## Summary

Closes #303. Adds four token-efficient `SAPWrite` actions so an LLM can edit a global ABAP class **without re-sending the full `/source/main`** (typically 1–80 lines instead of 500–4000):

| Action | Sends | Touches |
|---|---|---|
| `edit_class_definition` | new `CLASS … DEFINITION … ENDCLASS.` block | DEFINITION only (or a local include via `include=`) |
| `add_method` | a `METHODS` clause (+ optional `visibility`, `abstract`) | DEFINITION + IMPLEMENTATION atomically |
| `edit_method_signature` | new `METHODS` clause for one method | DEFINITION only (IMPL untouched) |
| `delete_method` | a method name | DEFINITION + IMPLEMENTATION atomically |

All four are backed by SAP's existing `/sap/bc/adt/oo/classes/{name}/objectstructure` endpoint — `parseClassStructure` returns the precise `#start=L,C;end=L,C` line ranges to splice, so **no client-side ABAP parsing of the existing source is needed** for the common path (abaplint stays as the diff-side declaration parser only).

### Design highlights
- **Refuse-policy** (`edit_class_definition`): a client-side diff against the current class structure refuses a write that would leave the class non-activatable — an added concrete method with no `METHOD…ENDMETHOD` body, or an orphan body after a removal — and points the caller at `add_method`/`delete_method`. `ABSTRACT METHODS`, `EVENTS`, `INTERFACES`, `ALIASES` are exempt.
- **Version consistency**: structure + source are fetched at the same effective version (active, or the inactive draft when one exists) and the version is threaded through the cache key, so chained surgery on a draft splices against aligned line ranges instead of silently corrupting it.
- **Cross-release**: NW 7.50 splits methods across `CLAS/OO` (def) + `CLAS/OM` (impl) merged by name; S/4HANA 2023 (kernel 7.58) uses one `CLAS/OM`. One parser handles both.
- **No auto-activate** — writes the inactive draft; caller runs `SAPActivate` next (matches existing `update`/`edit_method`).

### Verification
- **a4h (S/4HANA 2023, kernel 7.58)** — end-to-end live: 5/5 integration tests + a 12-step probe (`scripts/probe-issue-303.mjs`) PASS; the chained-draft scenario (`add_method` ×2 without activating between) confirmed both methods survive activation.
[...]

PR #316, opened 2026-05-29
feat(read): SAPRead grep — regex search within object source (#313)

marianfoo:
## Summary

Adds an optional **`grep="<regex>"`** parameter to `SAPRead`. Instead of returning a whole object's source, ARC-1 fetches it server-side, runs a case-insensitive regex, and returns **only the matching lines plus ±3 lines of context** (with line numbers). For `CLAS`, each match is annotated with the owning class/method so a hit leads straight to a targeted `method=` read.

This is the server-side **"search → locate → read"** loop agents already use with ripgrep on a local checkout — the one non-redundant capability from [issue #313](https://github.com/marianfoo/arc-1/issues/313). The surgery half of that suggestion already shipped in #307 (`edit_class_definition`, `add_method`, `method="*"`); this complements it. A 1,000-line full-class read for "where is `MARA` selected?" becomes a few hundred tokens of matches (Design Principle #3).

```
SAPRead(type="CLAS", name="ZCL_ORDER", grep="select.*from")  → matching lines, annotated by method
SAPRead(type="PROG", name="ZREPORT", grep="WRITE")           → matches + context, line numbers
```

## What changed

- **`src/context/grep.ts`** (new, pure, no I/O): `grepSource(source, pattern, {methods})` — `gim` regex with a **literal fallback** when the pattern isn't valid regex or has unescaped metacharacters (LLMs send `read_entities(` a lot), ±3 context, 100-match cap with a "narrow your pattern" note, and `[class=>method]` annotation. Mirrors `method-surgery.ts`; fully unit-tested in isolation.
- **`src/handlers/intent.ts`**: a `grepText` helper + `if (args.grep)` branch on every source-bearing read case — `PROG, CLAS, INTF, FUNC, INCL, DDLS, DCLS, BDEF, SRVD, DDLX, TABL, VIEW`. The `CLAS` branch reads the raw section, annotates via `listMethods`, rejects `grep`+`method`, and frames output with a `[name section=…]` header.
- **`src/adt/client.ts`**: `getClassInclude(name, include)` — raw section source (no `=== inc ===` wrapper that `getClass(name, include)` adds) so grep line numbers stay accurate.
- Three-file schema/tools/intent sync; docs (`tools.md`, `README`, `CLAUDE.md`, feature matrix, roadmap).

## Design notes

- **Reuses `MethodInfo.containingClass`** for annotation — no new field added to `method-surgery.ts` (it's untouched).
- **No new scope or SAP round-trip**: grep is a response transform over already-fetched, ETag-cached source under the existing `read` scope. (CLAS `include=` grep is the one uncached read, matching the existing non-grep include path.)
- `include="main"` correctly resolves to `/source/main`; a missing include returns a friendly message instead of a raw 404.

## Test plan

- **Unit** (full suite green, 3232 tests): `grep.ts` helper (16 tests — matching, context, separators, literal fallback, truncation, method annotation, between-method attribution), handler paths (PROG/BDEF/TABL/CLAS grep, no-match, invalid pattern, `grep`+`method` error, `include=main` routing, missing-include 404), schema/tools sync, `getClassInclude`.
- **E2E** (`tests/e2e/sapread-grep.e2e.test.ts`, runs via `npm run test:e2e`): greps the persistent `ZCL_ARC1_TEST` fixture — asserts filtered output, method annotation, and that grep output is smaller than the full read.
- `npm run typecheck` + `npm run lint` clean.

## Known limitation / follow-up

The pattern is a user-supplied `RegExp` run synchronously, so a pathological pattern (e.g. `(a+)+$`) against a long line could trigger catastrophic backtracking and block the event loop (a DoS vector on multi-user deployments). Mitigating it properly is a dependency/architecture decision (`re2` for linear-time matching, or a worker-thread timeout) rather than part of this feature — flagging for a follow-up. Input is authenticated and `read`-scoped in the meantime.


samibouge:
Seems to be working well, Claude uses grep all the time now when I'm working on code.

PR #335, opened 2026-06-03
fix(adt/client): copy tablWriteUrlCache in withSafety() clone

### LINKED ISSUE #333: fix(adt/client): copy tablWriteUrlCache in withSafety() clone (arc-1@0.9.7, possible regression)
`withSafety()` clones of `AdtClient` in arc-1@0.9.7 do **not** copy `tablWriteUrlCache`, so the clone has `tablWriteUrlCache = undefined`. This leads to a runtime error (TypeError on `.get()`) when TABL write/activate is invoked on a clone generated by `withSafety()`. 

> __Note__: We installed `arc-1@0.9.8` but haven't yet confirmed whether this bug is fixed there — please check before merging if newer releases supersede this fix.

---

**Trigger Condition:**
- The crash is triggered only when both `SAP_XSUAA_AUTH=true` and `extra.authInfo?.scopes` is present, which is the code path using `client.withSafety(effectiveSafety)` in `server.js`. Without these, the default client is used directly (constructor initializes `tablWriteUrlCache`).

---

**Fix / Minimal Diff (src/adt/client.ts):**
Add the following immediately after the `tablUrlCache` block in `withSafety()`:

```ts
Object.defineProperty(clone, 'tablWriteUrlCache', {
  value: this.tablWriteUrlCache,
  writable: false,
  enumerable: false,
});
```

**Original code for reference:**
```ts
withSafety(safety: SafetyConfig): AdtClient {
  const clone = Object.create(AdtClient.prototype) as AdtClient;
  Object.defineProperty(clone, 'http', { value: this.http, writable: false, enumerable: true });
  Object.defineProperty(clone, 'safety', { value: safety, writable: false, enumerable: true });
  Object.defineProperty(clone, 'username', { value: this.username, writable: false, enumerable: true });
  Object.defineProperty(clone, 'tablUrlCache', { value: this.tablUrlCache, writable: false, enumerable: false });
  //  ADD THIS BLOCK ↓↓↓
  Object.defineProperty(clone, 'tablWriteUrlCache', {
    value: this.tablWriteUrlCache,
    writable: false,
    enumerable: false,
  });
  //  END ADDITION ↑↑↑
  Object.defineProperty(clone, 'packageHierarchyResolverHolder', {
    value: this.packageHierarchyResolverHolder,
    writable: false,
    enumerable: false,
  });
  return clone;
}
```

**How to Reproduce:**
1. Set `SAP_XSUAA_AUTH=true` and supply `extra.authInfo?.scopes`
2. Use an ARC-1 server route or handler for tab write/activate (e.g. `handleSAPWrite`)
3. Observe the crash on missing cache with `TypeError: Cannot read properties of undefined (reading 'get')`

---

Assigning to @marianfoo for triage.
- If a newer version (≥0.9.8) fixes this, an upgrade is sufficient (please close as solved or link the PR).
- Otherwise, this is a one-line patch for continued support on v0.9.7.
- I can submit a PR for your preferred approach.

marianfoo:
## Summary

`AdtClient.withSafety()` builds the per-request safety clone with `Object.create(AdtClient.prototype)`, which **bypasses the constructor**, so every instance field must be re-attached by hand via `Object.defineProperty`. The clone was missing `tablWriteUrlCache`, leaving it `undefined` and crashing the first TABL write/activate with:

```
TypeError: Cannot read properties of undefined (reading 'get')
```

at `resolveTablObjectUrlForWrite()` (`this.tablWriteUrlCache.get(upper)`).

Closes #333.

## Root cause & regression

- `tablWriteUrlCache` + `resolveTablObjectUrlForWrite()` were introduced in #286 (v0.9.6), which also rerouted the TABL write/activate call sites from `resolveTablObjectUrl()` (uses the **copied** `tablUrlCache`) to `resolveTablObjectUrlForWrite()` (uses the **uncopied** `tablWriteUrlCache`) — but did not update `withSafety()`.
- Confirmed present in **v0.9.6, v0.9.7, and v0.9.8** — answering the reporter's open question: 0.9.8 does **not** fix it.
- Before #286 the equivalent paths used the copied read cache, so TABL writes on a clone worked → genuine regression, not just "possible."

## Trigger & impact

The clone is produced on every authenticated HTTP path (`src/server/server.ts`):

| Auth path | Affected |
|---|---|
| XSUAA OAuth / OIDC JWT (scopes) | ✅ |
| API key with a profile | ✅ (the issue's repro missed this second clone path) |
| stdio (no auth) | ❌ exempt — uses the constructor-built client (why CI never caught it) |

Affected operations: `SAPWrite` update/delete on TABL, `SAPActivate` single/batch on TABL — i.e. the common "create a table, then activate it" workflow was broken over authenticated connections.

## Fix

- Re-attach `tablWriteUrlCache` to the clone, **sharing the same `Map` instance** (matching `tablUrlCache`) — TABL/DT-vs-TABL/DS resolution is about object addressing, independent of per-request safety scope, so resolutions stay visible across the original and all clones.
- Add an `INVARIANT` guard-rail comment in `withSafety()` and a `CLAUDE.md` "Key Files" row so the next field added to `AdtClient` isn't forgotten (this was the second near-miss of this class).

## Tests

4 new unit tests in `tests/unit/adt/client.test.ts` (`describe('withSafety')`):
1. clone shares the same `tablWriteUrlCache` Map instance
2. clone shares the same `tablUrlCache` Map instance (closes the adjacent untested gap)
3. `resolveTablObjectUrlForWrite()` no longer throws on a clone (the regression)
4. a resolution cached on the original is visible from the clone (shared Map, not a copy)

Verified the 3 fix-dependent tests **fail without the fix** and pass with it.

- `npm test` → **3314 pass** (102 files)
- `npm run typecheck` → clean
- `npm run lint` → clean
- Codex review (`/codex:review`) → no blocking issues

No manual version bump — release-please will cut a patch from the `fix:` commit.


xkev1:
Hi Marain, I'm not able to test this in my system currently can you please push this in a next releasee so i can test this scenario directly

PR #339, opened 2026-06-03
feat(transport): set explicit transport target (TR_TARGET) on SAPTransport.create

marianfoo:
## Why

The first pass added a `transportLayer` param, but that's the wrong SAP field:

- **Transport layer** (Transportschicht) — a *package* property; influences the target only indirectly via the consolidation route.
- **Transport target** (Transportziel, data element `TR_TARGET`, SAP GUI field `KO013-TARSYSTEM`) — set *directly on the request*: a system (`C11`), system.client (`C11.021`), or group (`/TRG/`). The group / sys.client forms require **extended transport control (CTC)**.

## What

- **`SAPTransport(action="create", target="…")`** — sets `TR_TARGET` via the `tm:root`/`newrequest` endpoint (`POST /sap/bc/adt/cts/transportrequests`), the **only** ADT path that sets the target. The `CreateCorrectionRequest` endpoint silently ignores any target (verified). SAP validates the target server-side; an unknown target is converted into actionable guidance (valid forms + CTC requirement). An empty/whitespace `target` is rejected rather than silently falling through.
- **`transportLayer`** param on create → `?transportLayer=` (advanced layer override; abap-adt-api mechanism).
- **`layers`** discovery action (read scope) — lists the system's transport layers so a client picks a real value instead of guessing.
- The create response now reads the request back and **reports the resolved target**; an empty target ("Local Change Requests") = LOCAL request, surfaced with guidance.
- `parseTransportList` now extracts `target`/`targetDesc`.

## Verification

Live on **a4h (S/4HANA 2023, kernel 7.58)**:
- `target="LOCAL"` → request created; the `tm:root`/`newrequest` endpoint works on 7.58 (the "fails on NW 7.50" note is release-specific).
- `target="/TRG/"` / `"C11"` → **400 "Target 'X' does not exist"** (a4h has no transport routes), surfaced as friendly guidance. The happy path needs a CTC-enabled system — point the integration test at it with `TEST_TRANSPORT_TARGET=/TRG/`.
- `CreateCorrectionRequest` with a `<TARGET>` body element → ignored (stays local), confirming the endpoint split.

Unit + integration tests added (incl. a live "unknown target → clear error" integration test). Full unit suite: **3327 passing**; the only failures are the pre-existing `auth-rate-limit.test.ts` (stale `express-rate-limit` dep, unrelated to this change).

A Codex review pass was run on the diff; all five findings (empty-target fall-through, 404 handling, `&apos;` decoding, and two test-coverage gaps) are addressed in this PR.


marianfoo:
## NW 7.50 compatibility (requested test)

**Finding: the explicit-`target` path does NOT work on NW 7.50/7.51.** It uses `POST /sap/bc/adt/cts/transportrequests` with a `tm:root`/`newrequest` body — the exact path ARC-1's own `createTransport` JSDoc documents as *verified live on npl 7.50 SP02* to return **HTTP 400 "user action is not supported"** (`CL_ADT_TM_RESOURCE` ignores `tm:useraction` pre-7.52). It works on 7.52+/S4 (verified on a4h 7.58).

**Could not re-verify live:** npl.marianzeis.de (the 7.50 system) was down during this work — nginx returns 502, and it's hosted on a different box (176.9.72.62) than the one I have SSH access to (65.109.59.210, which only runs saprouter + a chrome container). So this rests on the documented prior live verification, not a fresh run.

**Fix added (`87ca60ed`):** the create handler previously only matched `/does not exist/i`, so a 7.50 rejection surfaced as a cryptic raw 400. It now catches "user action is not supported" and returns release guidance (needs NW 7.52+/S4; on older releases create the request without a target and set it in SE09/SE10). Covered by a unit test that mocks the 7.50 400. The tool `target` description notes the 7.52+ requirement.

The non-target paths (`createTransport` via `CreateCorrectionRequest`, `transportLayer`, target read-back) use the endpoint that ARC-1 documents as working on both 7.50 and 7.58. **Still worth a live 7.50 re-run once npl is back up** — especially the `layers` value-help and read-back paths.

marianfoo:
## ✅ Live-verified on NW 7.50 (npl 7.50 SP02)

npl is back up; ran the real ARC-1 code against it. Results:

| Path | NW 7.50 | Outcome |
|---|---|---|
| `createTransport` (no target) | ✅ works | created `NPLK900070`, read-back + delete OK (target/desc empty) |
| `createTransportWithTarget` (`target=…`) | ⛔ 400 | server returns `ADT_TM_COMMON_EXCEPTION` / **"user action is not supported"** — handler routes to the **release-guidance** branch (needs 7.52+) ✓ |
| `listTransportLayers` (`action="layers"`) | ⛔ 404 | value-help endpoint is **7.52+ only** ("No suitable resource found") |

**Two release gaps confirmed and both now handled gracefully** (verified live, not just mocked):
- `87ca60ed` — explicit `target` on 7.50/7.51 → friendly "needs NW 7.52+" guidance (the raw error is the cryptic "user action is not supported").
- `34638ee6` — `layers` on 7.50/7.51 → "discovery not available on this release" instead of a raw 404.

Net: on NW 7.50/7.51 the **no-target create path works** (CreateCorrectionRequest), and the two 7.52+-only features (`target`, `layers`) now degrade with clear, actionable messages rather than cryptic SAP errors. Full unit suite **3329 passing** (only the pre-existing `auth-rate-limit` dep failure remains).

marianfoo:
## Deep cross-source research → discovery-gating + official target value-help (`f089408a`)

Researched docs, repos, and SAP's own tooling, then verified live on both 7.50 and 7.58.

**Sources:** decompiled SAP Eclipse/adt-ls JARs (`com.sap.adt.tm.model`), SAP Notes 2046730/2047506/2558101/2982187/3720728, `abap-adt-api`/`sapcli`/`erpl-adt`, and live a4h 7.58 + npl 7.50.

**Key findings:**
- Two ADT resources: `/cts/transports` (CreateCorrectionRequest — 7.50-safe, no explicit target) vs `/cts/transportrequests` (`tm:root` — create-with-target, newer releases only). Confirmed ARC-1's split is correct.
- **No ADT way to set a target on 7.50, and no modify-target-on-existing anywhere** (`CL_ADT_CTS_MANAGEMENT` has only `CREATE_EMPTY_REQUEST`). Graceful degradation is the only correct option.
- **SAP's own Eclipse gates create-with-target on ADT *discovery capability*, not a version check** — the `transportorganizer.v1+xml` Accept type on `/cts/transportrequests`. Live: present on a4h 7.58, absent on npl 7.50. ARC-1's already-cached discovery map exposes it.
- There's an **official target value-help** (`/cts/transportrequests/valuehelp/target`) — the proper way to discover valid targets.

**Changes:**
- `supportsExplicitTransportTarget()` discovery-gate (Eclipse's exact mechanism, reusing the cached discovery map). Fails fast on 7.50 *before* a doomed POST; post-hoc "user action is not supported" catch kept as fallback.
- `targets` action (official value-help) — discover valid `target` values; gated the same way (NW 7.50 returns it as **HTTP 200 empty, not 404**, so gate on capability, not status).
- create's "does not exist" hint now points at `action="targets"`.

**Live verification:** a4h 7.58 — value-help returns `DEV`; `target="DEV"` → **201 with `tm:target=DEV` set** (the happy path, finally proven). npl 7.50 — gate → fail-fast, **no POST**. Full unit suite **3340 passing** (only the pre-existing `auth-rate-limit` dep failure).

PR #347, opened 2026-06-04
fix(probe): repair broken probe CLI and validate ABAP Platform 2025 (SAP_BASIS 816)

marianfoo:
## Summary

A third SAP test system — **ABAP Platform Trial 2025** — was added to the lab. Live probing established it reports **`SAP_BASIS 816`** (SAP renumbered the release scheme straight from the 7.5x line to 8.16 for 2025; releases 759–815 were consumed by quarterly S/4HANA Cloud Public Edition). This PR makes 816/2025 a **first-class, regression-locked test target** and fixes a real bug found while capturing its fixture.

Scoped to **validation + hardening** — *not* adding support for the 124 new ADT collections the 2025 system exposes (those are documented as future opportunities).

## What's in here

- **Fix the broken probe CLI.** `npm run probe` threw `discoveryMap.has is not a function`: `scripts/probe-adt-types.ts` treated `fetchDiscoveryDocument()`'s `{ map, nhiPresent }` return as a `Map`. Root cause — `scripts/` was excluded from `tsconfig.json`, so `tsc` never caught the API drift. Fixed the destructure, added `tsconfig.scripts.json`, and wired it into `npm run typecheck` (guard verified: reintroducing the bug now fails typecheck with 3 errors).
- **Commit the 2025 probe fixture** (`tests/fixtures/probe/abap-platform-2025-onprem-trial`, SAP_BASIS 816, S4FND 109, onprem, full RAP/CDS surface, 0 unavailable) + 3 replay tests that regression-lock it.
- **Make 8xx/816 release handling explicit + tested.** `src/` diff is **comment-only** — 816 is already handled correctly by monotonic numbering (`816 > 758`). Added the real-`816` assertions: post-stateful (no 423 lock-handle warning) and maps to abaplint `v758` (the correct on-prem baseline — `Cloud` would false-positive on classic on-prem ABAP).
- **Docs**: `SAP_ABAP_RELEASE` examples (CLAUDE.md, `.env.example`, configuration-reference), probe fixture inventory, 3-system research inventory, and the integration skip profile (abapGit ADT bridge absent on the trial; untuned-container write/activate note).

## Why `816 → v758` (not `Cloud`)

[...]

PR #350, opened 2026-06-05
fix(lint): don't block writes on parse errors beyond abaplint's grammar (8xx / SAP_BASIS 816)

marianfoo:
## Summary

Fixes a real bug found while researching SAP_BASIS **816** (ABAP Platform 2025) support: **arc-1's pre-write lint wrongly blocks writes of valid 816 syntax.**

arc-1 pins `@abaplint/core` to its highest grammar (`v758`) for any 8xx release. abaplint therefore can't parse legitimate new 816 constructs — CDS `define table entity`, `READ TABLE … WHERE`, new RAP/BDL keywords, etc. — and emits `parser_error` / `cds_parser_error`. The pre-write lint (`runPreWriteLint`) treats those as **blocking**, so `SAPWrite` of a valid 816 object is rejected before the PUT. abaplint's `Cloud` target fails identically, so mapping 8xx→Cloud is **not** a fix.

## Fix

When the detected SAP_BASIS release is beyond abaplint's grammar ceiling (`release > 758`), demote `parser_error` / `cds_parser_error` to **non-blocking warnings** in the pre-write config. SAP-side activation remains the definitive syntax check (consistent with how FUNC and unsupported types are already handled). On supported releases (`≤758`) genuine parse errors still block — no regression.

- `src/adt/features.ts`: `ABAPLINT_MAX_RELEASE` (= 758, kept in sync with the `mapSapReleaseToAbaplintVersion` ceiling) + `isBeyondAbaplintCeiling(release)`.
- `src/lint/config-builder.ts`: `buildPreWriteConfig` uses `Warning` severity for the parser-error rules when beyond the ceiling.

## Evidence (verified locally)

`validateBeforeWrite` on real 816 syntax, before vs after:

| Source | release | result |
|--------|---------|--------|
| `define table entity ZTE {…}` | 816 | **pass=true**, `cds_parser_error` → warning ✅ |
| `define table entity ZTE {…}` | 758 | pass=false, `cds_parser_error` (error) — still blocks ✅ |
| `READ TABLE itab … WHERE …` | 816 | **pass=true**, `parser_error` → warning ✅ |
| `READ TABLE itab … WHERE …` | 758 | pass=false, `parser_error` (error) — still blocks ✅ |

## Test plan

- `npm test` — 3381 pass (+7 new: `isBeyondAbaplintCeiling` in `features.test.ts`; demotion behavior in `lint.test.ts`)
- `npm run typecheck` — clean
- `npm run lint` — clean

## Notes

- **Re-bump path:** when abaplint ships a v759/816 grammar, raise `ABAPLINT_MAX_RELEASE` and the `mapSapReleaseToAbaplintVersion` ceiling together; the demotion then self-disables for releases abaplint can parse.
- Interim workaround that this replaces: `--lint-before-write=false`.
- Scope: on-prem 8xx (the verified case). BTP/Cloud has a related lag (abaplint `Cloud` is also behind) but its non-numeric release string isn't covered here — separate follow-up.
- Background: PR #347's `docs/research/abap-platform-2025-816-compatibility.md`, item 1.2.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -240,6 +240,7 @@ tests/
 | Add safety check | `src/adt/safety.ts` |
 | Add/modify PrettyPrint action | `src/adt/devtools.ts`, `src/handlers/intent.ts` (handleSAPLint), `src/handlers/tools.ts`, `src/handlers/schemas.ts` |
 | Add lint rule config | `src/lint/lint.ts`, `src/lint/config-builder.ts`, `src/lint/presets/` |
+| Adjust pre-write lint on releases beyond abaplint's grammar (8xx, e.g. SAP_BASIS 816 / ABAP Platform 2025) | `src/adt/features.ts` (`ABAPLINT_MAX_RELEASE`=758 + `isBeyondAbaplintCeiling`), `src/lint/config-builder.ts` (`buildPreWriteConfig` `parserSeverity`). abaplint v758 (its ceiling) can't parse new 816 syntax (CDS `define table entity`, `READ TABLE … WHERE`, RAP keywords) → would false-positive-block writes, so `parser_error`/`cds_parser_error` are **demoted to non-blocking warnings when `release > 758`** (activation is the definitive check); still block on ≤758. Bump `ABAPLINT_MAX_RELEASE` + the `mapSapReleaseToAbaplintVersion` ceiling together when abaplint adds a newer grammar. Tests: `tests/unit/{adt/features,lint/lint}.test.ts`. |
 | Add an ARC-1-native pre-write semantic hint for a new object type | `src/lint/pre-write-hints.ts` (add `inspect<Type>Source` → `LintResult[]` with `severity:'warning'`), `src/lint/lint.ts` (wire into `validateBeforeWrite()`, filename-gated by extension), `tests/unit/lint/{pre-write-hints,lint}.test.ts` |
 | Add dependency pattern | `src/context/deps.ts` |
 | Add CDS dependency pattern | `src/context/cds-deps.ts` |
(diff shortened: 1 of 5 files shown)


marianfoo:
Re-reviewed in detail + extended after a thorough self-review against the live 816 + 758 systems and arc-1's own linter:

**Verified correct & surgical** (`validateBeforeWrite`, on-prem):
| source | 816 | 758 |
|---|---|---|
| CDS `define table entity` | pass ✓ (warning) | blocks (error) |
| CDS `define external entity` | pass ✓ (warning) | blocks (error) |
| `READ TABLE … WHERE` | pass ✓ (warning) | blocks (error) |
| **valid classic ABAP** | pass, **0 warnings** | pass, 0 warnings |
| genuinely-broken ABAP | pass (warning, demoted) | blocks (error) |

Key point: valid code produces **no** spurious warning on 816 — only syntax abaplint truly can't parse triggers the demoted warning. 758 still blocks every parse error (no regression).

**Review finding + fix (new commit):** the same false-positive also hit the standalone `SAPLint action=lint` path (`buildLintConfig`), where 816 reported `parser_error` as a misleading (non-blocking) "error". Extended the release-gated demotion there too for consistency, + a coupling-guard test so `ABAPLINT_MAX_RELEASE` can't drift from the `mapSapReleaseToAbaplintVersion` ceiling.

**Merge-readiness:** #347 is now merged; #350 is `MERGEABLE` against it (the `features.ts` diffs are in different regions). `npm test` 3384 pass, typecheck + lint clean; all deterministic CI checks green (the `integration` check is the known live-SAP environmental flake, non-blocking).

PR #375, opened 2026-06-08
fix(write): emit pak:recordChanges so transportable package creation works on SAP_BASIS 816

### LINKED ISSUE #374: buildPackageXml omits pak:recordChanges → transportable package creation fails on SAP_BASIS 816 ("Change recording must be activated")
## Summary
`SAPManage(action="create_package")` always fails for a **transportable** (non-`$TMP`) package on SAP_BASIS 816 with `HTTP 400 — "Change recording must be activated for package <NAME>"`. `buildPackageXml()` (`src/adt/ddic-xml.ts`) emits `<pak:attributes pak:packageType="…"/>` and never sets `pak:recordChanges`, so the backend defaults change recording to `false` (KORRFLAG = space). Local `$TMP` packages work because recording is legitimately off there.

## Repro
1. Transportable parent package, non-local software component (e.g. `softwareComponent="HOME"`, a real `transportLayer`) on on-prem S/4HANA 816.
2. A modifiable Workbench transport (request- or task-id).
3. `SAPManage(action="create_package", name="Z…_CHILD", superPackage="Z…", softwareComponent="HOME", transportLayer="…", transport="<corrNr>")`
4. → `HTTP 400 "Change recording must be activated for package Z…_CHILD"`. Both corrNr variants fail identically.
5. The same package created via abapGit succeeds with KORRFLAG=X; arc-1 `$TMP` creation also works.

## Root cause
SAP_BASIS 816 enforces a constraint requiring KORRFLAG=X for transportable packages. The ADT field `record_changes` (serialized as the attribute `pak:recordChanges` on `<pak:attributes>`) maps 1:1 to KORRFLAG via `CL_PAK_ADT_PERSIST` with **no** auto-derivation from software component / transport layer — the backend honors exactly what the client sends. The corrNr is irrelevant (the check validates the package attribute, not the transport), which is why both request-id and task-id fail identically. The gap went unnoticed because the only package E2E test uses `$TMP`, and the `buildPackageXml` unit tests assert nothing about recording. (The upstream `abap-adt-api` `createBodyPackage()` has the same gap, which is where this derives from.)

### On-the-wire confirmation
Real captured ADT GET responses show the attribute on `<pak:attributes>`: a transportable package → `pak:recordChanges="true"`, a `$TMP` package → `pak:recordChanges="false"` (e.g. `DataZooDE/erpl-adt` captured fixtures; the attribute is also used by `jfilak/sapcli` and `abapify/adt-cli`, and an analogous fix shipped in `fr0ster/mcp-abap-adt-clients`).

## Suggested fix
Emit `pak:recordChanges` on the existing `<pak:attributes>` element. Add an optional `recordChanges?: boolean` to `PackageCreateParams` that defaults, when undefined, to a transportability heuristic: `true` when `softwareComponent` is set and not `LOCAL`, `false` for LOCAL/`$TMP` (so existing local creation is unchanged). PR to follow.

## Environment
On-prem S/4HANA, SAP_BASIS (ABAP) release 816.

abat-TAO:
Fixes #374

## What
`buildPackageXml()` now emits the ADT change-recording flag `pak:recordChanges` on the `<pak:attributes>` element for transportable packages. Adds an optional `recordChanges?: boolean` to `PackageCreateParams`; when omitted it defaults to a transportability heuristic — `true` for a non-`LOCAL` software component, `false` for LOCAL/`$TMP` (attribute omitted, exactly as today).

## Why
On SAP_BASIS 816 the backend enforces a constraint requiring KORRFLAG=X for transportable (non-LOCAL) packages. The ADT field `record_changes` (serialized as `pak:recordChanges`) maps 1:1 to KORRFLAG via `CL_PAK_ADT_PERSIST` with no auto-derivation, so a payload that omits it defaults recording OFF. `buildPackageXml` never emitted the flag, so every transportable `create_package` failed with `HTTP 400 "Change recording must be activated for package <NAME>"`. Local `$TMP` creation worked because recording is legitimately off there. (The same gap exists upstream in abap-adt-api; an analogous fix shipped in fr0ster/mcp-abap-adt-clients.)

## How
- Add `recordChanges?: boolean` to `PackageCreateParams` (an explicit value overrides the heuristic).
- In `buildPackageXml()`, compute `recordChanges = params.recordChanges ?? (softwareComponent set && !== 'LOCAL')` and append `pak:recordChanges="true"` to `<pak:attributes>` when on. LOCAL/`$TMP` is unchanged (attribute omitted → backend default false), so existing local behavior is untouched.

## Test plan
- `npm test` — added `buildPackageXml` unit tests (transportable → attribute present; LOCAL/`$TMP` → absent; explicit override). Updated the one existing `intent.test.ts` create_package assertion (it uses `softwareComponent="HOME"`, so it now expects `pak:recordChanges="true"`).
- `npm run typecheck` — clean (new optional field, no signature break).
- `npm run lint` — `biome check .` clean.
- Manual / integration: created a transportable sub-package under a non-local parent (softwareComponent=HOME) with a modifiable transport on an 816 system → succeeds with KORRFLAG=X; `$TMP` creation still works.

### Confirmation of the attribute
Real captured ADT GET responses show `pak:recordChanges` as an attribute on `<pak:attributes>` — `"true"` for a transportable package, `"false"` for `$TMP`.

### DIFF
--- a/src/adt/ddic-xml.ts
+++ b/src/adt/ddic-xml.ts
@@ -58,6 +58,13 @@ export interface PackageCreateParams {
   softwareComponent?: string;
   transportLayer?: string;
   packageType?: 'development' | 'structure' | 'main';
+  /**
+   * Whether the package records changes (ADT `record_changes` -> backend KORRFLAG).
+   * When omitted, defaults to `true` for transportable packages (a non-LOCAL software
+   * component) and `false` for LOCAL/$TMP packages. SAP_BASIS 816+ rejects a transportable
+   * package created with recording off ("Change recording must be activated for package ...").
+   */
+  recordChanges?: boolean;
 }
 
 export interface ServiceBindingCreateParams {
@@ -305,6 +312,12 @@ export function buildPackageXml(params: PackageCreateParams): string {
   const superPackage = params.superPackage ?? '';
   const softwareComponent = params.softwareComponent ?? 'LOCAL';
   const transportLayer = params.transportLayer ?? '';
+  // Transportable packages (a non-LOCAL software component) must record changes on
+  // SAP_BASIS 816+, otherwise the backend rejects creation with HTTP 400
+  // "Change recording must be activated for package <NAME>". LOCAL/$TMP keeps it off.
+  const isTransportable = softwareComponent.trim() !== '' && softwareComponent.trim().toUpperCase() !== 'LOCAL';
+  const recordChanges = params.recordChanges ?? isTransportable;
+  const recordChangesAttr = recordChanges ? ' pak:recordChanges="true"' : '';
 
   return `<?xml version="1.0" encoding="UTF-8"?>
 <pak:package xmlns:pak="http://www.sap.com/adt/packages"
@@ -315,7 +328,7 @@ export function buildPackageXml(params: PackageCreateParams): string {
              adtcore:version="active"
              adtcore:responsible="DEVELOPER">
   <adtcore:packageRef adtcore:name="${escapeXml(params.name)}"/>
-  <pak:attributes pak:packageType="${escapeXml(packageType)}"/>
+  <pak:attributes pak:packageType="${escapeXml(packageType)}"${recordChangesAttr}/>
   <pak:superPackage adtcore:name="${escapeXml(superPackage)}"/>
   <pak:applicationComponent/>
   <pak:transport>
--- a/src/adt/ddic-xml.ts
+++ b/src/adt/ddic-xml.ts
@@ -59,10 +59,9 @@ export interface PackageCreateParams {
   transportLayer?: string;
   packageType?: 'development' | 'structure' | 'main';
   /**
-   * Whether the package records changes (ADT `record_changes` -> backend KORRFLAG).
-   * When omitted, defaults to `true` for transportable packages (a non-LOCAL software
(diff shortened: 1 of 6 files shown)


marianfoo:
Pushed follow-up commit `9c3de8de` onto this PR branch.

What it adds on top of the original fix:
- exposes `recordChanges?: boolean` in `SAPManage(create_package)` schema/tool docs
- passes explicit `recordChanges` through the handler
- serializes `pak:recordChanges="true|false"` explicitly, including explicit false
- defaults false only for literal `LOCAL`; non-`LOCAL` components, including `ZLOCAL`, default true

Validation already run locally:
- `npm test -- tests/unit/adt/ddic-xml.test.ts tests/unit/handlers/intent.test.ts`
- `npm run typecheck`
- `npm run lint`
- `git diff --check`
- `npm test`

Live SAP checks already run on the same final tree:
- SAP_BASIS 758: package create under `ZLOCAL` with transport succeeded, readback showed `recordChanges="true"`, delete succeeded.
- SAP_BASIS 750: release/connectivity smoke passed; `/sap/bc/adt/packages` create endpoint returns ADT 404 `No suitable resource found` on that stack.

Closed duplicate #376 after moving the changes here.

abat-TAO:
Thanks for picking this up so fast and for the follow-up commit — much appreciated, glad it's merged! Happy to help again anytime.

PR #376, opened 2026-06-08
[codex] fix package recordChanges handling

### LINKED ISSUE #374: buildPackageXml omits pak:recordChanges → transportable package creation fails on SAP_BASIS 816 ("Change recording must be activated")

marianfoo:
Fixes #374.

## What

This updates `SAPManage(action="create_package")` and the package XML builder to set ADT package change recording explicitly via `pak:recordChanges`.

Compared with PR #375, this keeps the core fix for transportable packages and also:

- exposes `recordChanges?: boolean` in the public `SAPManage` schema and tool definition
- passes explicit `true` and `false` through the handler into `buildPackageXml()`
- defaults `pak:recordChanges="false"` only for the literal `LOCAL` software component with no transport layer
- defaults `pak:recordChanges="true"` for non-`LOCAL` software components, including `ZLOCAL`, and whenever a transport layer is supplied
- serializes `pak:recordChanges="true"` or `"false"` explicitly instead of relying on backend defaults

## Why

SAP_BASIS 816 rejects transportable package creation when the package is created without change recording active. Research showed `pak:recordChanges` is a first-class ADT package attribute, separate from the `corrNr` transport assignment.

Live S/4HANA 2023 evidence also showed that `ZLOCAL` must not be hard-coded as non-recording: on the test system, package `ZLOCAL` has `softwareComponent="ZLOCAL"`, `recordChanges="true"`, and requires a transport for child package creation. So the safe inferred default is only literal `LOCAL` => false; every named non-`LOCAL` component should record unless the caller explicitly overrides it.

## Tests

Local:

- `npm test -- tests/unit/adt/ddic-xml.test.ts tests/unit/handlers/intent.test.ts`
- `npm run typecheck`
- `npm run lint`
- `git diff --check`
- `npm test`

Live SAP checks:

- S/4HANA 2023 / SAP_BASIS 758: integration release smoke passed (`gets structured system info with user`, `gets installed components`).
- S/4HANA 2023 / SAP_BASIS 758: `SAPManage(create_package)` under `ZLOCAL` with an existing modifiable transport created a package, ADT readback returned `softwareComponent="ZLOCAL"` and `recordChanges="true"`, then `SAPManage(delete_package)` deleted it and a follow-up read returned 404.
- NetWeaver 7.50 / SAP_BASIS 750: integration release smoke passed (`gets structured system info with user`, `gets installed components`).
- NetWeaver 7.50 / SAP_BASIS 750: guarded `SAPManage(create_package)` returned ADT 404 `No suitable resource found` for `/sap/bc/adt/packages`; this confirms the package-create endpoint is not available on that 7.50 ADT stack, while general ADT connectivity works.

Note: `npm ci` was run before validation because the existing `node_modules` tree was out of sync with `package-lock.json`. The install completed with the existing Node engine warning (`current v22.21.1`, required `>=22.22.1`), and no dependency files changed.

### DIFF
--- a/src/adt/ddic-xml.ts
+++ b/src/adt/ddic-xml.ts
@@ -58,6 +58,12 @@ export interface PackageCreateParams {
   softwareComponent?: string;
   transportLayer?: string;
   packageType?: 'development' | 'structure' | 'main';
+  /**
+   * Whether the package records object changes in transport requests
+   * (`pak:recordChanges`, backend KORRFLAG). When omitted, ARC-1 infers it
+   * from transportability metadata and keeps literal LOCAL packages off.
+   */
+  recordChanges?: boolean;
 }
 
 export interface ServiceBindingCreateParams {
@@ -303,8 +309,11 @@ export function buildDataElementXml(params: DataElementCreateParams): string {
 export function buildPackageXml(params: PackageCreateParams): string {
   const packageType = params.packageType ?? 'development';
   const superPackage = params.superPackage ?? '';
-  const softwareComponent = params.softwareComponent ?? 'LOCAL';
-  const transportLayer = params.transportLayer ?? '';
+  const softwareComponent = params.softwareComponent?.trim() || 'LOCAL';
+  const transportLayer = params.transportLayer?.trim() ?? '';
+  const normalizedSoftwareComponent = softwareComponent.toUpperCase();
+  const isLocalSoftwareComponent = normalizedSoftwareComponent === 'LOCAL';
+  const recordChanges = params.recordChanges ?? (!isLocalSoftwareComponent || transportLayer !== '');
 
   return `<?xml version="1.0" encoding="UTF-8"?>
 <pak:package xmlns:pak="http://www.sap.com/adt/packages"
@@ -315,7 +324,7 @@ export function buildPackageXml(params: PackageCreateParams): string {
              adtcore:version="active"
              adtcore:responsible="DEVELOPER">
   <adtcore:packageRef adtcore:name="${escapeXml(params.name)}"/>
-  <pak:attributes pak:packageType="${escapeXml(packageType)}"/>
+  <pak:attributes pak:packageType="${escapeXml(packageType)}" pak:recordChanges="${boolToXml(recordChanges)}"/>
   <pak:superPackage adtcore:name="${escapeXml(superPackage)}"/>
   <pak:applicationComponent/>
   <pak:transport>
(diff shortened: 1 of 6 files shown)


marianfoo:
Moved the follow-up changes onto #375 so the original PR/author stays as the active contribution. Closing this duplicate PR.

PR #384, opened 2026-06-09
feat(multi-system): single ARC-1 instance routes to N SAP systems via ARC1_SYSTEMS

Prolls:
## Motivation

Working with multiple SAP systems from a single AI session is a common workflow:
- **Compare code across systems** — e.g. check how a program differs between production and development without switching MCP connections
- **Read reference data from pre-production** — e.g. fetch real table contents from a sandbox to write accurate unit test fixtures, without having to maintain two separate MCP server configurations

Before this PR, each system required its own ARC-1 deployment and a separate MCP server entry in the client config. This PR collapses N deployments into one.

## Summary

Adds multi-system support so a single ARC-1 deployment can serve multiple SAP systems. Each tool call can optionally include a `system` parameter to route to a specific system; omitting it routes to the primary system as before.

### How it works

Configure secondary systems via `ARC1_SYSTEMS`:

```
ARC1_SYSTEMS="prod:MY_PROD_DEST:MY_PROD_PP_DEST:developer,sandbox:MY_SBX_DEST:MY_SBX_PP_DEST:viewer-data"
```

Format: `alias:dest_basic:dest_pp[:profile]` — comma-separated, profile defaults to `developer`.

At startup, ARC-1 builds a shared client per system by resolving the `dest_basic` BTP destination (including Cloud Connector proxy for on-premise). At request time, PP clients are built using `dest_pp` with the per-system proxy — so on-premise systems work correctly with Cloud Connector.

When `ARC1_SYSTEMS` is set, the `system` enum is automatically injected into every tool's input schema, listing the available aliases.

### Safety

Each system entry carries a `profile` that acts as an additional safety ceiling, intersected with the user's scopes at request time. This means a secondary system can be locked to `viewer-data` even though the server-wide config allows writes.

### Changes

- **`src/server/types.ts`** — `SystemEntry` interface + `systems: SystemEntry[]` in `ServerConfig`
- **`src/server/config.ts`** — `parseSystems()` + `ARC1_SYSTEMS` env var parsing
- **`src/server/server.ts`** — system client pool (built at startup), per-request routing, per-system PP proxy, system-profile safety ceiling
- **`src/handlers/schemas.ts`** — `system?: string` added to all 16 tool schemas
- **`src/handlers/tools.ts`** — `system` enum injected into tool definitions when `ARC1_SYSTEMS` is set
- **`tests/unit/server/config.test.ts`** — 8 tests for `parseSystems`
- **`manifest-multi.yml`** — example CF manifest for a multi-system deployment

### Key fix included

On-premise secondary systems require a Cloud Connector proxy to be reached from BTP CF. The initial implementation passed `undefined` as the proxy for secondary PP clients, causing `AdtNetworkError: fetch failed`. The fix stores the per-system proxy in the client pool and uses it when building PP clients at request time.

## Test plan

- [ ] `npm test` — 3488/3496 pass (7 pre-existing failures on `main`, unrelated to this PR)
- [ ] `npm run typecheck` — clean
- [ ] Manual: deploy with `ARC1_SYSTEMS="sys1:dest1:pp1:developer,sys2:dest2:pp2:viewer-data"` — startup logs show both systems registered
- [ ] Manual: tool call without `system` → routes to primary system
- [ ] Manual: tool call with `system="sys1"` → routes to sys1, PP auth uses sys1's proxy
- [ ] Manual: tool call with `system="unknown"` → clean error listing available systems
- [ ] Manual: on-premise secondary system — PP auth succeeds and ADT calls go through Cloud Connector

### DIFF
--- a/manifest-multi.yml
+++ b/manifest-multi.yml
@@ -1,15 +1,15 @@
 ---
-# Cloud Foundry manifest — ARC-1 multi-system (DS7 + VS7) instance.
+# Cloud Foundry manifest — ARC-1 multi-system instance.
 #
-# Replaces the two separate arc1-ds7 / arc1-vs7 deployments with a single app
-# that routes tool calls to the right system via the `system` parameter.
+# A single ARC-1 deployment that routes tool calls to N SAP systems via the
+# `system` parameter. Each system gets its own safety profile ceiling.
 #
-# Primary system (SAP_BTP_DESTINATION) = DS7 (developer profile, writes enabled)
-# Secondary system via ARC1_SYSTEMS    = VS7 (viewer-data profile, read-only)
+# Primary system (SAP_BTP_DESTINATION)  = routes when `system` is omitted
+# Secondary systems via ARC1_SYSTEMS    = routed when `system=<alias>` is passed
 #
 # ARC1_SYSTEMS format: alias:dest_basic:dest_pp[:profile],...
-#   - dest_basic = BasicAuthentication destination (used at startup to build the shared client)
-#   - dest_pp    = PrincipalPropagation destination  (used at request time for PP)
+#   - dest_basic = BasicAuthentication destination (shared client built at startup)
+#   - dest_pp    = PrincipalPropagation destination  (per-user, built at request time)
 #   - profile    = safety ceiling for that system (viewer|viewer-data|developer|admin|...)
 #
 # To deploy:
@@ -17,7 +17,7 @@
 #   cf push -f manifest-multi.yml
 #
 # To set secrets (do NOT commit these):
-#   cf set-env arc1-multi ARC1_API_KEYS "<key>:admin"
+#   cf set-env <app-name> ARC1_API_KEYS "<key>:admin"
 #
 applications:
   - name: arc1-multi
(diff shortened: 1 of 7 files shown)


marianfoo:
Thank you for your PR but it is the architecture decision to have one ARC-1 instance per SAP System as this would increase complexity and decrease security.

marianfoo:
I had a few similar requests like this. Please review https://github.com/arc-mcp/mcp-hub if this could help you. So far this is only usable in BTP CF.
Would that be helpful and what is your enviornment?

Prolls:
Hi @marianfoo, thanks for pointing me to mcp-hub. After spending time reading through the code, I understand why my PR was rejected; the architecture is cleaner with real isolation between systems and no changes to the arc-1 core.

That said, I want to flag a concrete constraint that mcp-hub doesn't address for my setup, in case it's useful feedback for the roadmap.

BTP CF Quota
My subaccount has tight resource limits (RAM, number of apps, service instances) and 10+ sap environments in my landscape. With mcp-hub, a 3-system setup requires 4 CF apps (~900 MB RAM), 4 XSUAA service instances, and 4 routes. My original PR kept it to 1 app (~256 MB) and 1 XSUAA instance. On a trial subaccount or a small BTP contract, that difference is the line between "deployable" and "not deployable."

I am not asking you to change the architecture decision, but I think this trade-off is worth documenting explicitly in the mcp-hub README under "When NOT to use it":

Tight CF quota: mcp-hub requires one CF app per SAP system plus the hub itself. If your subaccount has strict RAM or app-count limits, this may not be viable. In that case, a single arc-1 instance with a read-only secondary system configured via environment variables is a reasonable workaround (at the cost of weaker isolation).

Thanks again for building this—the /all aggregation approach in particular is a nicer UX than what I had.

marianfoo:
Please add your feedback here to make this possible: https://github.com/arc-mcp/arc-1/issues/577

PR #392, opened 2026-06-10
fix: harden SAPRead grep regex handling

marianfoo:
## Goal

Close security register item R2 by reducing ReDoS risk in the optional `SAPRead.grep` server-side source filter.

## Changes

- Adds shared grep limits for pattern length and per-line search input.
- Rejects high-risk regex constructs before compiling: lookaround assertions, backreferences, and nested quantified groups.
- Exposes the same max length in the Zod schemas and MCP tool JSON schema.
- Keeps rendered grep output based on original source lines while matching only bounded line prefixes.

## Review

Re-reviewed the scoped diff before publishing. The change is limited to grep runtime/schema/tool surfaces and focused unit coverage. It does not change any ADT endpoint behavior.

## Validation

- `npm test -- tests/unit/context/grep.test.ts tests/unit/handlers/schemas.test.ts tests/unit/handlers/tools.test.ts` passed, 274 tests
- `npm run typecheck` passed

Live SAP validation was not run; this is a local post-fetch formatter hardening and the worktree has no `INFRASTRUCTURE.md` credentials/recipes.

### DIFF
--- a/src/context/grep.ts
+++ b/src/context/grep.ts
@@ -37,13 +37,19 @@ export interface GrepResult {
   matchCount: number;
   /** Formatted, LLM-friendly match report (or a no-match / invalid-pattern message). */
   output: string;
-  /** True only when the pattern is not valid regex AND has no literal match either. */
+  /** True when the pattern cannot be used safely or has no valid regex/literal interpretation. */
   invalidPattern: boolean;
 }
 
 const DEFAULT_CONTEXT_LINES = 3;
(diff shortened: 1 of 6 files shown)


marianfoo (commented):
Reviewed the grep hardening, pushed 5b7740bf for quantified alternation ReDoS patterns, and reran focused checks. Ready from my side; GitHub won’t let me self-approve.

PR #393, opened 2026-06-10
fix: isolate PP cache state

marianfoo:
## Goal

Close security register items R1, R3, and R12 by preventing shared cache state from crossing user authorization boundaries under principal propagation.

## Changes

- Keys inactive-object list caching by the authenticated user key when PP is active.
- Bypasses inactive-list caching when no non-empty user key is available.
- Routes inactive-list invalidation through the same effective key helper.
- Disables dependency payload cache read/write for `SAPContext` under PP.
- Disables `SAPContext(action="usages")` under PP because it reads the shared warmup reverse-dependency index, and points callers to live `SAPNavigate(action="references")` instead.

## Review

Re-reviewed the scoped diff before publishing. The branch contains only cache-isolation changes and tests; the overlapping SRVB publish/unpublish package-gate hunks were intentionally removed from this PR.

## Validation

- `npm test -- tests/unit/cache/inactive-list-cache.test.ts tests/unit/cache/caching-layer.test.ts tests/unit/handlers/intent.test.ts` passed, 742 tests
- `npm run typecheck` passed

Live SAP validation was not run because the worktree has no `INFRASTRUCTURE.md` credentials/recipes.

### DIFF
--- a/src/cache/inactive-list-cache.ts
+++ b/src/cache/inactive-list-cache.ts
@@ -2,49 +2,53 @@ import type { AdtClient } from '../adt/client.js';
 import type { InactiveObject } from '../adt/types.js';
 
 interface CachedInactiveList {
-  username: string;
+  userKey: string;
   objects: InactiveObject[];
   fetchedAt: number;
 }
 
 export class InactiveListCache {
-  private byUsername = new Map<string, CachedInactiveList>();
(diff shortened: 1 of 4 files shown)


marianfoo:
Reviewed cache-isolation changes and pushed follow-up hardening for PP inactive-list keys. Checks pass; approved from my side (GitHub blocks formal self-approval).

PR #394, opened 2026-06-10
fix: enforce SRVB package gate

marianfoo:
## Goal

Close security register item R4 by enforcing the package allowlist before `SAPActivate(action="publish_srvb")` and `SAPActivate(action="unpublish_srvb")` perform mutating SRVB publish/unpublish requests.

## Changes

- Resolves the real SRVB package from `/sap/bc/adt/businessservices/bindings/{name}` using the service-binding v2 metadata accept type.
- Reuses the existing fail-closed `enforceAllowedPackageForObjectUrl` helper.
- Blocks restricted-package publish/unpublish before `/publishjobs` or `/unpublishjobs` is reached.
- Keeps the unrestricted-package fast path unchanged.

## Review

Re-reviewed the scoped diff before publishing. The branch only adds the SRVB package gate and focused tests; it does not include the cache-isolation changes from the overlapping handler file.

## Validation

- `npm test -- tests/unit/handlers/intent.test.ts` passed, 705 tests
- `npm run typecheck` passed

Live SAP validation was not run because the worktree has no `INFRASTRUCTURE.md` credentials/recipes. Endpoint/package metadata behavior was checked against the local Eclipse ADT/discovery references during implementation.

### DIFF
--- a/src/handlers/intent.ts
+++ b/src/handlers/intent.ts
@@ -5933,6 +5933,12 @@ async function handleSAPActivate(
     if (!name) {
       return errorResult('Missing required "name" parameter for publish_srvb action.');
     }
+    await enforceAllowedPackageForObjectUrl(
+      client,
+      objectUrlForType('SRVB', name),
+      `Publish of service binding '${name}'`,
+      SERVICEBINDING_V2_CONTENT_TYPE,
+    );
     const serviceType = await resolveServiceType();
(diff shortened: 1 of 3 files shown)


marianfoo (commented):
Reviewed the SRVB package gate and focused tests; no changes from me. Looks good, but GitHub blocks self-approval from this account.

PR #411, opened 2026-06-12
fix(context): avoid SAPContext dependency convoy

marianfoo:
## Summary

Fixes external feedback point 2: `SAPContext` dependency resolution no longer waits on fixed five-request waves. It now uses a bounded worker pool that keeps up to five dependency fetches in flight and starts the next dependency as soon as a slot frees up.

## Changes

- Add ordered `mapWithConcurrency()` helper in `src/context/compressor.ts`.
- Use it for ABAP contract dependency fetches and CDS dependency fetches.
- Add deterministic unit tests proving the sixth dependency starts after the first worker slot frees, not after the full first batch completes.
- Add completed ralphex plan: `docs/plans/completed/ralphex-context-worker-pool.md`.

## Validation

- `npm test -- tests/unit/context/compressor.test.ts`
- `npm test`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
  - `SAPRead(type="SYSTEM")` succeeded through built CLI
  - `SAPRead(type="COMPONENTS")` confirmed `SAP_BASIS=758`, `S4FND=108`

## Live SAP Notes

Local credentials for NW 7.50 and A4H 2025 were not present in this workspace (`NPL_*` and A4H 2025 env vars unset). This change is release-invariant: it schedules the same existing ADT calls without changing URLs, headers, payloads, parsing, or safety behavior. Repository PR CI is labeled `A4H 2025` and is the available 2025 live verification path.

### DIFF
--- a/docs/plans/completed/ralphex-context-worker-pool.md
+++ b/docs/plans/completed/ralphex-context-worker-pool.md
@@ -1,0 +1,95 @@
+# Ralphex Plan: SAPContext Dependency Worker Pool
+
+## Overview
+
+This plan fixes the convoy effect in `SAPContext` dependency resolution. The existing implementation resolves dependencies in fixed five-request waves; the next wave cannot start until the slowest request in the current wave completes. The target behavior is a bounded worker pool that keeps up to five dependency fetches in flight and starts the next dependency as soon as any slot frees up.
+
+The change is intentionally local to context compression. It does not add an ADT endpoint, alter safety policy, change response schemas, or change dependency ordering in the final result.
+
+## Context
+
+### Current State
+
+`fetchContractsParallel()` in `src/context/compressor.ts` slices dependencies into `MAX_CONCURRENT` batches and awaits `Promise.all()` for each batch. `resolveCdsDepthLevel()` uses the same fixed-wave pattern for CDS dependency reads. Heterogeneous SAP request latency means four worker slots can sit idle behind one slow dependency.
+
+The repository already has a FIFO semaphore in `src/adt/semaphore.ts`, but this call site only needs bounded ordered mapping inside `compressor.ts`; pulling in the ADT semaphore would not remove much complexity because the result array must remain dependency-order stable.
+
+### Target State
+
+`SAPContext` dependency resolution uses a small local `mapWithConcurrency()` helper. It preserves result order, caps concurrency at `MAX_CONCURRENT`, and starts the next dependency immediately after any current dependency settles.
+
+### Key Files
+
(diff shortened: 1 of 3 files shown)


marianfoo:
## Deep-PR review — ✅ APPROVE

Independent review (re-verified from sources, not from the PR's claims). The bounded worker pool correctly removes the fixed-wave convoy without changing observable behaviour.

### Verified independently
- **Order preserved.** `mapWithConcurrency` writes results by claimed `index`, so output order matches input order regardless of completion order — the property `resolveCdsDepthLevel`/`fetchContractsParallel` callers rely on.
- **Concurrency capped correctly.** `workerCount = min(concurrency, items.length)`; each worker claims an index with a synchronous `nextIndex++` *before* the first `await`, so two workers can never claim the same index (safe in single-threaded JS). At most `MAX_CONCURRENT` mappers run at once.
- **Failure semantics are identical** — this is the part that mattered most, and it holds: both mappers swallow their own errors and never reject. `fetchSingleContract` returns a `success:false` contract on `catch` (`src/context/compressor.ts:185-194`); `fetchCdsDependency` does the same (`src/context/compressor.ts:457-466`). So the old `Promise.all`-per-wave vs new worker-pool difference in rejection handling is moot in practice.
- **No new ADT endpoint / safety / schema / scope surface** — only scheduling of existing client calls changes.
- **Gates (run locally on this branch):** `typecheck` ✅ · `lint` ✅ · `npm test` ✅ **3725 passed**.

### Tests
The two added tests (ABAP + CDS) genuinely distinguish old vs new: under the previous fixed-wave code the 6th dependency cannot start until all of waves 1–5 settle, so `waitUntil(() => started.includes('…DEP6'))` would time out. I re-ran both **5×in isolation → 5/5 pass**, and they also passed under 3-way parallel suite contention, so the `setTimeout`/`Date.now()` budget isn't marginal.

### Optional nits (non-blocking, the helper is otherwise correct)
- **`concurrency <= 0` footgun.** With `concurrency <= 0`, `workerCount` becomes 0, no workers spawn, and the function returns an array of `undefined` holes silently. Not reachable today (callers pass the `MAX_CONCURRENT = 5` constant), but a `Math.max(1, …)` clamp or assert would harden the reusable helper.
- **Post-failure drain.** *If* a mapper could ever reject, `Promise.all` would reject while the surviving workers keep draining the remaining queue in the background. Harmless today because mappers don't reject — worth a one-line comment stating that invariant so a future mapper change doesn't quietly regress it.

No changes required to merge.

PR #412, opened 2026-06-12
fix(xml): reuse parser array tag set

marianfoo:
## Summary

Fixes external feedback point 1: the `fast-xml-parser` `isArray` callback no longer allocates a repeatable-tag array and runs `.includes()` for every parsed XML tag.

## Changes

- Hoist repeatable ADT tag names into a module-level `Set`.
- Change `isArray` to `ARRAY_TAGS.has(name)`.
- Add direct `parseXml()` regression coverage for repeatable single tags and non-repeatable single tags.
- Add completed ralphex plan: `docs/plans/completed/ralphex-xml-array-tags-set.md`.

## Validation

- `npm test -- tests/unit/adt/xml-parser.test.ts`
- `npm test`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
  - `SAPRead(type="COMPONENTS")` succeeded and confirmed `SAP_BASIS=758`, `S4FND=108`

## Live SAP Notes

Local credentials for NW 7.50 and A4H 2025 were not present in this workspace (`NPL_*` and A4H 2025 env vars unset). This change is release-invariant: it preserves parser options and XML semantics, changing only the lookup structure used by `isArray`. Repository PR CI is labeled `A4H 2025` and is the available 2025 live verification path.


marianfoo:
## Deep-PR review — ✅ APPROVE

Independent review (re-verified from sources, not from the PR's claims). This is a pure, behaviour-preserving optimisation of the ADT XML parse hot path.

### Verified independently
- **Tag set is byte-identical.** Sorted diff of the old inline `isArray` list (HEAD) vs the new `ARRAY_TAGS` set (this branch): **37 ↔ 37 tags, zero difference**, no hidden duplicates (`sort -u` count == raw count on both sides). `Set.has(name)` is semantically equivalent to `Array.includes(name)` for string keys, so the parser contract is unchanged.
- **No security/architectural surface touched** — only `src/adt/xml-parser.ts`; no ADT endpoint, safety guard, schema, scope, or `console.log`.
- **Gates (run locally on this branch):** `typecheck` ✅ · `lint` ✅ (biome, no fixes) · `npm test` ✅ **3725 passed**.

### Tests
The two added `parseXml()` cases (single repeatable → array, single non-repeatable → object) lock both sides of the contract. Adequate.

### Optional nit (non-blocking)
- The new tests exercise a representative pair of tags, not the full list. If you want a guardrail against a future accidental tag drop, a tiny test asserting a couple of the less-obvious entries (e.g. `SEU_ADT_REPOSITORY_OBJ_NODE`, `orglvlinfo`) still array-ify would help — but the equivalence is otherwise self-evident.

Solid win. No changes required.

PR #413, opened 2026-06-12
perf(lint): memoize abaplint default configs

marianfoo:
## Summary

Fixes external feedback point 3: repeated `Config.getDefault(version)` construction is now memoized while keeping `Registry` instances fresh per parse.

## Changes

- Add `src/lint/abaplint-config-cache.ts` with:
  - `getDefaultAbaplintConfig(version)` for shared default configs
  - `cloneDefaultAbaplintConfig(version)` for mutable config-builder callers
  - test-only cache clear helper
- Route context parser and lint call sites through the cache.
- Keep `new Registry(config)` per parse so file/parse state is never shared.
- Add cache identity and clone-isolation tests.
- Add completed ralphex plan: `docs/plans/completed/ralphex-abaplint-config-memo.md`.

## Validation

- `npm test -- tests/unit/lint/abaplint-config-cache.test.ts tests/unit/lint/config-builder.test.ts tests/unit/lint/lint.test.ts tests/unit/context/deps.test.ts tests/unit/context/contract.test.ts tests/unit/context/method-surgery.test.ts`
- `npm test`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
  - `SAPRead(type="COMPONENTS")` succeeded and confirmed `SAP_BASIS=758`, `S4FND=108`

## Live SAP Notes

Local credentials for NW 7.50 and A4H 2025 were not present in this workspace (`NPL_*` and A4H 2025 env vars unset). This change is release-invariant: it affects local abaplint config construction only and does not alter ADT calls, headers, payloads, parsing of SAP responses, or safety behavior. Repository PR CI is labeled `A4H 2025` and is the available 2025 live verification path.

### DIFF
--- a/docs/plans/completed/ralphex-abaplint-config-memo.md
+++ b/docs/plans/completed/ralphex-abaplint-config-memo.md
@@ -1,0 +1,111 @@
+# Ralphex Plan: Memoize abaplint Default Config
+
+## Overview
+
+This plan addresses external feedback point 3: several hot paths call `Config.getDefault(version)` during every parse or lint-config build. The `Registry` must stay fresh because it holds file/parse state, but the default abaplint config is version-constant and can be memoized.
+
+The implementation adds a small cache module under `src/lint/` and routes all production `Config.getDefault()` call sites through it. Parser users still construct a new `Registry(config)` per parse. Config-builder users receive an isolated JSON clone before applying presets or user overrides, so no mutable caller can corrupt the cached default.
+
+## Context
+
+### Current State
+
+`Config.getDefault()` is called in dependency extraction, contract extraction, method surgery, lint config building, pre-write config building, and the legacy lint fallback. On workflows like `SAPContext`, a single request can parse the root source plus many dependencies, rebuilding the same default config repeatedly.
+
+### Target State
+
+Default configs are memoized by abaplint `Version`. Registry instances remain per-call. Config-builder code gets fresh clones derived from a cached serialized default config, preserving caller isolation while avoiding repeated default-rule construction.
+
+### Key Files
+
+| File | Role |
+|------|------|
+| `src/lint/abaplint-config-cache.ts` | Central cache for abaplint default configs and cloneable default JSON |
+| `src/context/deps.ts` | AST dependency extraction for `SAPContext` |
+| `src/context/contract.ts` | Public contract extraction for classes/interfaces |
+| `src/context/method-surgery.ts` | Method list/extract/splice AST parsing |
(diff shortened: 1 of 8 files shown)


marianfoo:
## Deep-PR review — ✅ APPROVE

Independent review (re-verified from sources, not from the PR's claims). Memoising the version-keyed default config is safe because abaplint's `Config` is immutable in practice, and the clone path keeps mutable callers isolated.

### Verified independently (the key question: is sharing one `Config` across many `Registry` parses safe?)
- **abaplint treats `Config` as immutable.** `node_modules/@abaplint/core/build/src/config.js:7` literally states `// assumption: this class is immutable`. All `this.config` mutations live in the **constructor** only; there are no post-construction setters that parsing invokes.
- **`Registry` never mutates the config.** It stores it by reference (`this.conf = conf`, `registry.js:157`) and only *reads* it via `getGlobal()` / `getVersion()` (`registry.js:184,193,312`). So sharing the same instance across `new Registry(config)` calls in `deps.ts` / `contract.ts` / `method-surgery.ts` is safe.
- **`get()` returns the internal object by reference** (`config.js:105-107`) — the one real sharing hazard. But the only `.get()` consumer is `cloneDefaultAbaplintConfig`, which merely `JSON.stringify`s it (read) and hands back a fresh `JSON.parse` clone, so `buildLintConfig`/`buildPreWriteConfig` mutate an isolated copy. The direct-use callers never call `.get()`.
- **`lint.ts` `DEFAULT_CONFIG` (now the shared v702 instance) is never mutated** — only passed to `new Registry(…)` (`lint.ts:65,91`). And it was *already* a module-level shared singleton before this PR, so the "shared config across parses" pattern is pre-existing and proven, not introduced here.
- **All production `Config.getDefault()` sites are rerouted** — `grep` confirms the only remaining occurrence in `src/` is the doc comment in `config-builder.ts`.
- **No ADT / safety / schema / scope surface** — internal lint/context only.
- **Gates (run locally on this branch):** `typecheck` ✅ · `lint` ✅ · `npm test` ✅ **3726 passed** (incl. the new cache test file).

### Tests
The three cache tests cover same-version identity, cross-version separation, and clone isolation — good.

### Optional nit (non-blocking)
- There's no explicit test asserting that *parsing through the shared config doesn't mutate it*. It's covered indirectly (the `deps` / `contract` / `method-surgery` / `lint` / `config-builder` suites all run on the shared path and pass), but a small regression test — parse a file via `getDefaultAbaplintConfig(v)`, then assert `cfg.get()` deep-equals a fresh `Config.getDefault(v).get()` — would lock the immutability assumption against a future abaplint upgrade that breaks it.

No changes required to merge.

PR #440, opened 2026-06-15
fix(http): handle 304/204/205 null-body status on Cloud Connector proxy path

ClementRingot:
## TL;DR

On Cloud Connector / RISE deployments, **any write whose read-before-write revalidates a cached object crashes** when SAP returns `304 Not Modified`. The connectivity-proxy code path reconstructs an HTTP `Response` by hand with a non-null body on a null-body status, which the Fetch spec forbids — so the constructor throws. One-line semantic fix + regression test.

## Symptom (observed in the field)

S/4 on-prem reached via BTP Cloud Connector + Principal Propagation. `SAPWrite action=edit_method` fails with:

```
ADT network error: Response constructor: Invalid response status code 304
Hint: Cannot reach the SAP system. Run SAPRead(type="SYSTEM") once as a connectivity probe before retrying...
```

The hint is misleading — **the SAP system was up the whole time.** Logs also showed many `304`s, which led to the root cause.

## Root-cause analysis

1. ARC-1 caches source by `(type, name, version)` with the SAP **ETag** and revalidates via conditional GET (`If-None-Match`). Unchanged object → SAP answers **`304`**. Expected, healthy; lots of 304 = good cache-hit rate.

2. `edit_method` does a **read-before-write**: it GETs the class source to splice the method. The object is cached, so that read is a conditional GET → **304**.

3. On Cloud Connector / RISE, every ADT call goes through `doProxyRequest()` (`src/adt/http.ts`), which does **not** use `fetch()` — it uses `undici.Client` and **manually reconstructs** a `Response`:

   ```ts
   const responseBody = await resp.body.text();   // '' for a 304 — empty string, NOT null
   return new Response(responseBody, { status: resp.statusCode /* 304 */, headers });
   ```

   Per the WHATWG Fetch spec, **204 / 205 / 304** are *null-body statuses*: the `Response` constructor **throws** if given a non-null body. `''` is non-null → `TypeError: Response constructor: Invalid response status code 304`, surfaced as `AdtNetworkError`.

### Why it was never seen before
The direct `fetch()` path (local stdio / direct HTTP) returns a native 304 `Response` and never reconstructs one, so 304 flows through fine. The bug is **specific to the Cloud Connector / RISE proxy path**.

### Why the 304s and the write error are the same root cause
The 304s in the logs are the read-cache revalidation; the write failure is that same 304 crashing the hand-rolled `Response` during the read-before-write. One problem, not two.

## The fix

`src/adt/http.ts` — pass a `null` body for null-body statuses:

```ts
const isNullBodyStatus = resp.statusCode === 204 || resp.statusCode === 205 || resp.statusCode === 304;
return new Response(isNullBodyStatus ? null : responseBody, { status: resp.statusCode, headers: responseHeaders });
```

The 304 then reaches `handleResponse()` normally (`< 400`, so not an error), `fetchSource()` sets `notModified: true`, and the caching layer serves the cached source — the write proceeds.

## Scope / blast radius

- Only touches the **proxy reconstruction** path (`doProxyRequest`). The direct `fetch()` path is unchanged.
- No behavior change for 200 / 4xx / 5xx.
- Restores correctness **and** keeps the ETag cache benefit (no need for the `ARC1_CACHE=none` workaround).

## Tests

Regression test added in `tests/unit/adt/http.test.ts` (`proxy configuration` block), driving a 304 through the proxy path:
- **Without the fix:** fails with the exact field error (`AdtNetworkError: … Invalid response status code 304`).
- **With the fix:** passes (`statusCode === 304`, body `''`).

`http.test.ts` (121 passing), `npm run typecheck`, and `biome check` all green.

## Operator workaround (pre-fix)

`ARC1_CACHE=none` stops sending `If-None-Match` (no 304, no crash) at the cost of re-downloading sources every call. The fix removes the need for it.

---
Validated live on a Cloud Connector sandbox deployment before opening this PR.

### DIFF
--- a/src/adt/http.ts
+++ b/src/adt/http.ts
@@ -1238,6 +1238,8 @@ export class AdtHttpClient {
       // yields '' not null). Pass null so a 304 from a conditional GET (ETag
       // revalidation) survives the proxy path instead of crashing the write
       // (e.g. edit_method's read-before-write through the Cloud Connector).
+      // (1xx informational statuses are null-body too, but never surface here:
+      // undici's Client doesn't return them as a final status and ADT never emits them.)
       const isNullBodyStatus = resp.statusCode === 204 || resp.statusCode === 205 || resp.statusCode === 304;
       return new Response(isNullBodyStatus ? null : responseBody, {
         status: resp.statusCode,
--- a/src/adt/http.ts
+++ b/src/adt/http.ts
@@ -1233,7 +1233,13 @@ export class AdtHttpClient {
         }
       }
 
-      return new Response(responseBody, {
+      // Statuses 204/205/304 are "null body status" per the Fetch spec — the
+      // Response constructor throws if given a non-null body (and `.text()`
+      // yields '' not null). Pass null so a 304 from a conditional GET (ETag
+      // revalidation) survives the proxy path instead of crashing the write
+      // (e.g. edit_method's read-before-write through the Cloud Connector).
+      const isNullBodyStatus = resp.statusCode === 204 || resp.statusCode === 205 || resp.statusCode === 304;
+      return new Response(isNullBodyStatus ? null : responseBody, {
         status: resp.statusCode,
         headers: responseHeaders,
       });
--- a/tests/unit/adt/http.test.ts
+++ b/tests/unit/adt/http.test.ts
@@ -1368,6 +1368,27 @@ describe('AdtHttpClient', () => {
       // The path should be the full URL (standard HTTP proxy protocol)
       expect(clientRequestPath(0)).toBe('http://sap.example.com:8000/path?sap-client=001&sap-language=EN');
     });
+
+    it('handles a 304 (null-body status) from the proxy without crashing the Response constructor', async () => {
+      // Regression: a conditional GET (If-None-Match) that revalidates to 304
+      // over the Cloud Connector proxy must not throw "Invalid response status
+      // code 304" — that crashed edit_method's read-before-write on RISE/CC.
+      mockClientRequest.mockResolvedValueOnce(mockClientResponse(304, '', { etag: '"abc"' }));
+
+      const client = new AdtHttpClient({
+        ...getDefaultConfig(),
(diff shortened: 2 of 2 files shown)


marianfoo:
## Reviewed in depth — good to merge ✅

Thanks @ClementRingot. The diagnosis in the description is exactly right, and I verified the fix independently rather than from the write-up.

**Reproduced the bug live** on a Cloud Connector (`ProxyType=OnPremise`) deployment: reading the same object twice makes the second read a conditional GET (`If-None-Match`) that revalidates to `304`, and on the proxy path that crashes with the exact `Response constructor: Invalid response status code 304` — surfaced as `AdtNetworkError` with the misleading "Cannot reach the SAP system" hint while SAP was up the whole time.

**Local full-suite run** (fork PRs don't get the full CI here, so I ran the gates locally on the rebased branch `62a5fde`):

| Gate | Result |
|------|--------|
| `npm ci` | exit 0, **0 vulnerabilities** |
| `npm run typecheck` | ✅ pass |
| `npm run lint` | ✅ pass (2 pre-existing biome infos) |
| `npm test` | ✅ **126 files / 3841 tests passed** |

**Regression test proven both ways:** reverting only `src/adt/http.ts` to the parent commit (keeping the new test) makes it **fail** with `AdtNetworkError: … Response constructor: Invalid response status code 304`; restoring the fix makes it **pass** (`statusCode === 304`, `body === ''`). So the test genuinely guards the regression, and it's in the right `proxy configuration` block.

**Blast radius** is ideal: only `doProxyRequest()`'s manual reconstruction changes — the direct `fetch()` path and 200/4xx/5xx behavior are untouched, `responseBody` is drained before the null-body branch (no socket leak), and it's the only `new Response(` in `src/`. Bonus: this also hardens any `204`/`205` returned by writes/deletes/unlock over the proxy, which would have crashed identically.

marianfoo:
Thank you! Had the problem reported before but did know how to fix

PR #508, opened 2026-06-25
feat: SAPQuery datapreview metrics + SAPDiagnose ABAP trace requests

marianfoo:
## Summary

Two features for debugging slow OData / SQL on large CDS stacks (the "serial-number search is very slow" use case).

### 1. SAPQuery — surface datapreview metrics
The ADT `datapreview/freestyle` response already carries `totalRows`, `queryExecutionTime`, and `executedQueryString`; ARC-1 parsed only `{columns, rows}` and dropped the rest. SAPQuery now returns them first:

```json
{ "totalRows": 318837, "queryExecutionTimeMs": 21.846, "executedQueryString": "...", "rowsReturned": 2000, "columns": [...], "rows": [...] }
```

`totalRows` is the real match count (independent of the row cap) and `queryExecutionTimeMs` is server-side time — exactly the "847 ms, N rows matched" signal needed for perf triage. Implemented as a dedicated `client.runQueryWithMetrics` so the metrics (incl. the internal generated SQL) surface **only** for SAPQuery, not for every `runQuery` consumer.

### 2. SAPDiagnose — arm/list/cancel ABAP trace requests
Three new actions close the loop on profiler tracing (the existing `action=traces analysis=dbAccesses` could only *read* a trace that someone else recorded):

- **`trace_start`** — arm a profiler trace (2 POSTs: `.../abaptraces/parameters` then `.../requests`); defaults to the connected user's next HTTP/URL request with SQL tracing on.
- **`trace_requests`** — list armed requests.
- **`trace_cancel`** — delete an armed request.

Flow: arm → reproduce the slow OData call → read it via `SAPDiagnose(action="traces", analysis="dbAccesses")` to see which tables/joins dominate. Arming/cancelling are `write` / `OperationType.Update`; listing is `read`.

## Live verification (a4h — S/4HANA 2023, SAP_BASIS 758)
- SAPQuery metrics confirmed (incl. a 318,837-row scan → `queryExecutionTimeMs: 21.846`, which settles the unit as **milliseconds**).
- Trace `start → list → cancel` round-trip verified end-to-end with clean cleanup (no armed request left behind).
- Trace REST endpoints are discovery-confirmed identical on 7.50 and 8.16 (release-invariant).

## Reviews
Passed a Claude `/code-review` pass and a Codex review. Hardening applied:
- **`trace_cancel` arbitrary-DELETE (HIGH):** the id→path guard now validates on the **decoded + normalized** path and requires a single slash-free segment under `abaptraces/requests/`, so neither `..` nor encoded separators (`%2f`/`%2e`) can escape — while still sending the original `%2c`-encoded id SAP expects. Tests cover both traversals.
- **`runQuery` spillover:** narrowed `runQuery`; metrics live in `runQueryWithMetrics` only (no leak into `SAPRead type=SOBJ` etc.); also removes the chunked path's N× double-parse.
- Guards for `expiresHours:0`, empty `<totalRows/>`, and empty-`id` create responses.

## Notes
- New tool surface grows the SAPDiagnose schema; tool-definition snapshots refreshed and schema/file-size budgets bumped deliberately.
- A known footgun (an `http` trace for the connected user can be consumed by ARC-1's own ADT traffic) is documented in the runtime hint and the dossier; a URL-pattern object filter to avoid it is noted as future work.
- Full unit suite green (the one failing `ui.test.ts` is pre-existing on `main` — unbuilt UI assets, unrelated).

Research dossier: `docs/research/abap-trace-requests-and-sapquery-metrics.md` · Plan: `docs/plans/sapquery-metrics-and-abap-trace-requests.md`


marianfoo:
Review pass completed and pushed to `feat/sapquery-metrics-trace-requests`.

Fixes added:
- Hardened `trace_cancel` id normalization against literal, encoded, and double-encoded path separators/traversal, backslashes, absolute non-request paths, and blank ids before DELETE.
- Changed `trace_start` `expiresHours=0` from silent 24h defaulting to an explicit validation error.
- Tightened datapreview metric parsing so blank numeric fields cannot become spurious zeroes.
- Added tests proving chunked SAPQuery stays on the narrow `runQuery` shape and does not surface metrics/executed SQL from chunk responses.
- Added safety-ceiling tests showing trace request create/cancel are blocked when `allowWrites=false`.

Gate:
- `npm test`: only the known clean-main failure in `tests/unit/server/ui.test.ts` (static asset fallback expected 200, got 404); all other unit tests passed.
- `npx vitest run tests/unit/adt/diagnostics.test.ts tests/unit/adt/xml-parser.test.ts tests/unit/handlers/search-navigate.test.ts`: passed.
- `npm run typecheck`: passed.
- `npm run lint`: passed, with existing Biome config info messages.
- `npm run validate:policy`: passed.
- `npm run build`: passed.
- `node scripts/ci/check-file-sizes.mjs`: passed.

PR #543, opened 2026-07-03
Multi-backend support in one instance

lemaiwo:


marianfoo:
## Proposed direction: endpoint-bound, strict-PP, read-only multi-target v1

First, thank you for building and testing this. The PR proves that one ARC-1 process can host multiple destination-backed runtimes, and the destination registry, lazy resolution, route dispatch, audit context, and destination-keyed feature isolation give us valuable working groundwork.

After a deeper security and operations review, we would like to discuss a narrower first supported mode. This is a discussion proposal, not yet a request for further implementation or a final specification.

### Goal

Support approximately 20 SAP systems with about three clients each in one ARC-1 deployment, without adding a target parameter to any MCP tool. Each SAP system/client is a separate MCP URL:

```text
https://arc1.example/A4H/100/mcp
https://arc1.example/A4H/200/mcp
https://arc1.example/A4P/100/mcp
```

The path binds the target before tool dispatch. There is no bare/default `/mcp` route in this mode and no way for tool arguments to select or override the target.

### Proposed v1 boundary

- Cloud Foundry HTTP deployment only.
- On-premise SAP through Destination Service, Connectivity Service, and Cloud Connector only.
- XSUAA-authenticated human users only; corporate IdP federation through XSUAA is supported.
- Strict principal propagation for every SAP request.
- Exactly one `OnPremise` + `PrincipalPropagation` destination per system/client.
- No shared SAP credentials or fallback identity.
- Read-only tools only. Writes, activation, transport writes, Git writes, plugin execution, and other mutations are structurally unavailable.
- Source reads are always available subject to SAP authorization. Data preview and free SQL are explicit system-level opt-ins.
- Cache mode `none`; warmup, UI, and plugins are disabled.
- Client `000` is rejected.
- API-key authentication is rejected because an API key has no human JWT that can be propagated.
- Direct raw Entra/IAS OIDC bearer tokens and cloud-to-cloud destinations are deferred.
- ATC and ABAP Unit are not redesigned as part of this feature; operators can use the existing general deny-action configuration if needed.

### Configuration: one `targets.yaml`

The current PR distributes configuration between `SAP_BTP_DESTINATIONS`, global MTA variables, destination `arc1.*` properties, and suffixed variables such as `SAP_ALLOWED_PACKAGES_S4D`. That is flexible, but at 60 targets it becomes difficult to review and typo-prone. Package policies also have no purpose in this read-only v1.

We propose one strict, secret-free inventory file:

```yaml
version: 1
mode: multi-target-readonly

systems:
  A4H:
    policy:
      allowDataPreview: true
      allowFreeSql: true
    clients:
      "100":
        destination: ARC1_A4H_100_PP
      "200":
        destination: ARC1_A4H_200_PP

  A4Q:
    policy:
      allowDataPreview: true
      allowFreeSql: false
    clients:
      "100":
        destination: ARC1_A4Q_100_PP
      "200":
        destination: ARC1_A4Q_200_PP

  A4P:
    policy:
      allowDataPreview: false
      allowFreeSql: false
    clients:
      "100":
        destination: ARC1_A4P_100_PP
```

The schema is intentionally narrow. It contains only routing and the two supported read-policy opt-ins. It does not contain environment labels, display metadata, URLs, credentials, authentication settings, write/package policies, feature overrides, client-specific policy overrides, or arbitrary environment variables.

System route IDs are uppercase alphanumeric. Client keys are exactly three digits. The client key is authoritative: `/A4H/100/mcp` always causes ARC-1 to add `sap-client=100` to ADT requests. The referenced destination must also explicitly contain `sap-client=100`; missing or mismatched values quarantine that target.

The presence of `ARC1_TARGETS_FILE` selects multi-target mode and activates the restrictions before the file is parsed. The required `mode: multi-target-readonly` value is an additional human-readable/schema assertion, not the only security switch.

The base `mta.yaml` would contain an inactive example:

```yaml
properties:
  # Multi-target strict-PP mode (read-only v1):
  # 1. Copy config/targets.example.yaml to config/targets.yaml
  # 2. Configure one PP destination per system/client with matching sap-client
  # 3. Uncomment this property, then rebuild and deploy the MTAR
  # ARC1_TARGETS_FILE: config/targets.yaml
```

Users build their deployment-specific MTAR from source. `targets.yaml` is packaged in the application. Cockpit therefore shows only the `ARC1_TARGETS_FILE` environment variable, not an editable file. Inventory or policy changes require rebuild/redeploy/restart.

### One destination, not a BasicAuth + PP pair

The current single-system MTA recommends a BasicAuth destination for startup feature probes/warmup and a second PP destination for user requests. Multi-target v1 deliberately defers that pattern.

Each target has only its PP destination. ARC-1 can retrieve and validate the destination configuration without retrieving a user authentication token or calling SAP. Actual SAP access is created only after an authenticated request supplies the XSUAA user JWT. Feature discovery is conservative/lazy and must not require a shared startup identity.

The immutable destination identity consists of URL, proxy type, authentication type, `sap-client`, and Cloud Connector location. These fields are pinned until restart. Credentials and user tokens may rotate, but identity drift quarantines only that target instead of silently retargeting an endpoint.

### Authenticated endpoint directory

We propose an authenticated `GET /` directory. It reads only the static validated inventory and never calls Destination Service or SAP. Any XSUAA user with ARC-1 `read` scope can see the configured system/client URLs; this is an inventory directory, not a claim that the user has SAP access to each target.

It should expose only system ID, client, and public MCP URL—not destination names, policies, credentials, target errors, or SAP health. Unknown routes still return a generic `404` without listing targets.

For example:

```text
A4H / 100  https://arc1.example/A4H/100/mcp
A4H / 200  https://arc1.example/A4H/200/mcp
A4P / 100  https://arc1.example/A4P/100/mcp
```

The page can also generate a VS Code / GitHub Copilot `.vscode/mcp.json` sample:

```json
{
  "servers": {
    "arc1-a4h-100": {
      "type": "http",
      "url": "https://arc1.example/A4H/100/mcp"
    },
    "arc1-a4h-200": {
      "type": "http",
      "url": "https://arc1.example/A4H/200/mcp"
    },
    "arc1-a4p-100": {
      "type": "http",
      "url": "https://arc1.example/A4P/100/mcp"
    }
  }
}
```

VS Code can perform OAuth discovery/authentication for HTTP MCP URLs; no API key should appear in the sample. One caveat: a normal browser address-bar request does not carry a bearer token. A bearer-authenticated JSON/HTML directory is feasible in the backend; a seamless interactive browser login would require a minimal AppRouter/session surface and can remain deferred with the rest of the UI.

### Startup, failure, and audit behavior

- Invalid YAML, unknown keys, duplicate/case-colliding routes, invalid clients, forbidden global options, or shared credentials fail complete startup.
- A missing, unreachable, incorrectly typed, or client-mismatched destination quarantines only its target and returns generic `503`; other targets stay available.
- Run at least two CF application instances.
- Add a global concurrency cap, a default per-target cap, and bounded queue wait so one slow SAP system cannot starve all targets.
- Configuration is restart-immutable. Destination identity drift requires restart; transient connectivity failures can retry with bounded backoff.
- Unknown endpoints never reveal the configured target list.
- Add durable, redacted `mcp_auth_rejected`, `pp_access_denied`, and `sap_access_denied` audit events with request ID, system/client, destination alias, user/subject identity, and stable reason code. Never record JWTs, assertions, cookies, destination tokens, or SAP response bodies.

### Main differences from the current PR

| Area | Current PR | Proposed v1 |
|---|---|---|
| Target identity | Destination name | Logical system + client |
| URL | `/mcp/<destination>` plus default `/mcp` | `/<SYSTEM>/<CLIENT>/mcp`; no default |
| Inventory | `SAP_BTP_DESTINATIONS` CSV | Strict `targets.yaml` |
| Policy | Global ceiling ∩ destination `arc1.*` ∩ suffixed env pins | Mandatory system policy with only data/SQL opt-ins |
| Authentication | Shared API key/OIDC/XSUAA modes | XSUAA human JWT only |
| SAP identity | Shared destination or optional `arc1.pp_destination` | Strict PP, exactly one destination per target |
| Writes | Can be enabled per destination | Structurally rejected |
| Cache/warmup | Per-destination cache and warmup supported | Cache none; warmup off |
| Feature discovery | Destination-keyed probing | Conservative/lazy without shared SAP identity |
| Unknown route | May return configured destination names | Generic `404` |
| Target failure | Lazy initialization/error response | Target-local quarantine with generic external error |
| Root discovery | No authenticated system/client directory | Authenticated inventory + generated VS Code sample |

We expect significant parts of the tested PR to remain reusable: destination registry lifecycle, exact route dispatch, lazy target runtime creation, Destination/Connectivity proxy handling, target audit context, and destination-keyed feature/discovery state. The proposal mainly narrows configuration, identity, safety, and failure semantics.

### Explicitly deferred and documented as unsupported

- Writes, activation, transport/Git mutations, and write-related package policies.
- A BasicAuth/startup-probe destination paired with each PP destination.
- API keys and non-user tokens.
- Direct Entra/IAS OIDC bearer propagation.
- Cloud-to-cloud destinations.
- stdio multi-target mode.
- Client `000`.
- Per-client policy differences.
- Cache, warmup, UI, and plugins.
- Live configuration reload.

### Questions for discussion

1. Does adapting the tested registry from destination-key routes to logical system/client routes look compatible with the current implementation, or is there a hidden dependency on exposing the destination name in the URL?
2. Can the existing destination lookup obtain configuration-only metadata for a `PrincipalPropagation` destination without creating a shared SAP client, or should we add a narrow configuration-only Destination Service lookup?
3. Are there any PP behaviors from your live test that require `arc1.pp_destination` to remain separate from the listed destination?
4. Which registry, route-dispatch, proxy, audit-context, and feature-isolation pieces do you think can be retained with minimal change under these constraints?
5. Do you see a client compatibility issue with `/<SYSTEM>/<CLIENT>/mcp` or with having no bare `/mcp` fallback?

If we agree on the direction, the next step would be a focused implementation plan and documentation update rather than expanding this discussion comment further.

Useful references:

- [SAP BTP principal propagation](https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/principal-propagation)
- [Destination Service “Find a Destination” response structure](https://help.sap.com/docs/r/cca91383641e40ffbe03bdc78f00f681/Cloud/en-US/83a3f3b9cd314618aba651044ed5b9df.html)
- [SAP MTA parameters and properties](https://help.sap.com/docs/btp/sap-business-technology-platform/parameters-and-properties)
- [VS Code MCP server configuration](https://code.visualstudio.com/docs/agent-customization/mcp-servers)

lemaiwo:
Thanks for the thorough review. Before the individual points, one framing remark, because I think we're using "security boundary" for two different things.

My position: multi-backend ARC-1 should behave exactly like single-backend ARC-1, per target, and the security bar should be an equivalence guarantee, not a feature cut. Concretely: one instance serving N backends must be exactly as secure as N separately deployed instances. Everything that is allowed on a standalone instance today is allowed on a target (as the same deliberate opt-in), and everything that is isolated between two separate instances must be provably isolated between two targets: SAP sessions, credentials, caches, feature state, guardrail ceilings, audit trails. That second half is the genuinely new security work in multi-target, and the tested PR already implements most of it (per-target cache files, destination-keyed feature/discovery state, per-target ceilings, per-user narrowing on top).

What I don't agree with is deriving security from removing capabilities. ARC-1 is already deny-by-default — writes, data preview, free SQL, transport and Git mutations are all off unless an operator explicitly enables them. That means a fresh multi-target deployment with no opt-ins already is your read-only posture, without a special mode. Structurally rejecting writes doesn't close a vulnerability we've identified — it just guarantees a fork of the feature the moment someone needs AI-assisted development on their DEV system, which is the primary use case. The same applies to cache none, XSUAA-only, and single-destination-only: on a single system these are all operator choices today; making them per-target choices keeps one consistent model instead of two ARC-1s that behave differently depending on how many backends are configured.

Where I fully agree — and these I'd treat as hard requirements for multi-target regardless of feature set:

- strict per-target isolation (the equivalence guarantee above)
- destination identity pinning and quarantine-on-drift, target-local failures with generic external errors
- no enumeration: unknown routes return a generic 404 (the current PR's 404 lists configured names — that goes)
- the durable redacted audit events (mcp_auth_rejected, pp_access_denied, sap_access_denied) with target context
- global + per-target concurrency caps so one slow system can't starve the rest
- PP strict semantics per target: where an operator runs strict PP, non-JWT callers (API keys) are rejected — which is existing single-system behavior, not a new rule

Regarding your questions:

1. Hidden dependency on the destination name in the URL? — No hidden dependency: the route segment is just the registry key. But I would not use system/client as the route either — see below.
2. Configuration-only destination lookup without creating a SAP client? — Yes, definitively — it already exists and the PR uses it. The lookup calls the Destination Service "find destination" endpoint with a client-credentials token only: no user JWT, no SAP contact, no SAP client created. The full configuration (URL, auth type, proxy type, sap-client, Cloud Connector location, custom properties) comes back as metadata. No new lookup needs to be built.
3. Does anything require arc1.pp_destination to stay separate? — Yes, when cache/warmup/startup-probe are in use: the BasicAuth destination provides the technical identity those need, the PP destination carries the per-user requests. That's the standard pair from the single-system setup, and under the equivalence principle it should remain available per target for operators who use those features. A target that runs strict PP without cache/probe needs only the PP destination — both configurations fall out of the same model, no special casing.
4. What can be retained? — Proxy handling (per-destination Cloud Connector incl. location ID) and the feature-isolation mechanism: unchanged. Registry and route dispatch: kept, with the pinning/quarantine state machine added (today a failed init retries on the next request — quarantine is an addition, not a rewrite). Audit context: kept; your event types plug into the existing typed union. Feature probing under strict PP is the one piece needing redesign — the probe currently uses the shared technical client, so for PP-only targets it must run lazily with the first authenticated user's session or stay conservative.
5. Client compatibility with /<SYSTEM>/<CLIENT>/mcp or no bare /mcp? — No client-side issue with deep paths. Bare /mcp I'd keep only when exactly one backend is configured (no wrong system to hit, and existing client configs keep working); with two or more it should 404.

My remarks on the configuration surface:

**Inventory belongs on the destinations, not in a packaged yaml.** The MTA configuration should focus on building and deploying. Your own schema hints at the redundancy: it requires sap-client in both the yaml and the destination, with quarantine on mismatch — because the destination already carries everything. I'd flip it: destinations opt in via an explicit marker property (arc1.expose=mcp), the same idiom Mobile Services and BAS use — no marker, not picked up. The MTA configuration becomes identical for every deployment, and onboarding a system is a cockpit/IaC action instead of a rebuild. This is also a security improvement: destination administration can be limited via BTP Role Collections to a small set of administrators, whereas anyone with CF deploy rights can ship a rebuilt MTAR containing a modified targets.yaml. For the audit trail, destinations can be managed as code (BTP Terraform provider, btp CLI), ARC-1 logs the effective inventory at startup, and your authenticated GET / directory is the human-readable view. To keep your immutability guarantees: discovery runs at startup only in v1 — adding a system is a restart, not a redeploy — identity stays pinned exactly as you propose, and a live reload endpoint stays deferred.

**Route segment: the destination name.** System/client in the path is what forces the yaml into existence. The destination name is already a unique, admin-chosen, charset-validated identifier — and it's what the tested PR routes on. If /A4H/100-style semantics matter, a naming convention (A4H_100) achieves it, with sap-client still taken authoritatively from the destination property. An optional arc1.path property could give prettier paths later, but only as part of the pinned identity set: changing it after registration quarantines the target, a collision quarantines both — so "no silent retargeting" holds.

**Guardrails on the destination, within the existing model.** Same three-layer model as the current PR: deny-by-default baseline, per-target opt-ins via arc1.* properties (Role-Collection-controlled), per-caller narrowing via scopes/profiles on top. An operator who wants your exact v1 posture simply enables nothing — and can additionally pin targets read-only at deploy time if cockpit-side changes must not matter.

**One engine, no modes.** I see no reason to differentiate the server logic between one backend and many — single-backend is the N=1 case of the same registry, with everything isolated per target. That also keeps the upgrade path linear: a customer who starts read-only and later enables writes on DEV changes one property, not their deployment architecture.

In the end this simplifies the setup to: an MTA configuration that is always the same, one destination per backend carrying connection + opt-in marker + guardrails, and security coming from the equivalence guarantee plus ARC-1's existing deny-by-default model — identical whether you run one backend or twenty.

marianfoo:
Thanks again for the implementation and discussion. After the feedback and discussions, I think the destination-based approach is the right direction, but I would like to adjust the architecture before merging it.

Proposed v1 design:

- Multi-target is explicit and default-off via `ARC1_MULTI_TARGET=true`.
- ARC-1 discovers subaccount destinations marked with `arc1.expose=mcp` once during startup.
- Each SAP SID/client gets a fixed endpoint:
  - `/A4H/100/mcp`
  - `/A4H/200/mcp`
  - `/PRD/100/mcp`
- No tool receives a system, client, or destination parameter. The endpoint fixes the target before the MCP request is processed.
- `arc1.system` must be a real SAP SID: exactly three uppercase alphanumeric characters, starting with a letter. `sap-client` is mandatory and exactly three digits.
- Discovered targets never automatically receive `/mcp`. Bare `/mcp` exists only when an old-style single target is explicitly configured. Legacy `/mcp` and named routes may coexist, even with different policies; ARC-1 will warn if the same destination appears through both.
- V1 uses one on-premise PrincipalPropagation destination per target. No second BasicAuth destination is required and there is no shared-user fallback.
- Destination changes become active only after restarting the CF app.
- Starting with zero targets is valid: `/health` remains healthy, ARC-1 logs a warning, and destinations can be configured afterward.

Minimum destination configuration is read-only:

```properties
Name=ARC1_A4H_100_PP
Type=HTTP
URL=http://a4h-abap:50000
ProxyType=OnPremise
Authentication=PrincipalPropagation
sap-client=100
arc1.expose=mcp
arc1.system=A4H
```

Optional target properties include:

- `arc1.allow_writes`
- `arc1.allow_data_preview`
- `arc1.allow_free_sql`
- `arc1.allow_transport_writes`
- `arc1.allow_git_writes`
- `arc1.allowed_packages`
- `arc1.allowed_transports`
- `arc1.deny_actions`
- `arc1.max_concurrent`

The effective policy is:

`global MTA/env ceiling ∩ destination policy ∩ XSUAA user scopes ∩ SAP authorization`

Target capability booleans default to false. Therefore data preview, SQL, and writes must be enabled explicitly for that destination and must also be permitted by the global ARC-1 ceiling. Existing write/package/transport/Git checks remain unchanged. If target packages are omitted, the global package ceiling—including the current `$TMP` default—still applies.

Configuration is fail-closed:

- Unknown `arc1.*` properties quarantine only that destination.
- Duplicate destination names quarantine every duplicate.
- Multiple destinations claiming the same SID/client route quarantine every claimant.
- Conflicted or invalid targets do not appear in routing or the catalog.
- External errors never enumerate targets; actionable details remain in redacted operator logs.
- Per-user destination lookups must still match the startup URL, client, authentication type, proxy type, and Cloud Connector location. Drift is rejected until restart.

An authenticated `/` and `/targets` catalog will be available to the existing `read` role. It will show only target labels, MCP URLs, and a VS Code/GitHub Copilot `mcp.json` example—not destination names, SAP URLs, policies, credentials, or conflicts. Per-target ARC-1 ACLs are beyond v1; SAP authorization through PP remains the final user-specific access check.

For v1, cache is forced to `none`; UI, plugins, ATC, and ABAP Unit are disabled for discovered targets. S/4HANA Public Cloud, SaaS/subscriber discovery, hot reload, and target-specific XSUAA roles remain follow-ups.

I opened draft PR #579 based on this PR and current `main`. This preserves your commits and attribution while giving us a separate current-main branch for the larger redesign and BTP integration/E2E testing. Nothing from the current implementation has to be retained if the new structure can be simpler or safer.

The detailed implementation and acceptance plan is in PR #579. Could you please review the architecture there, especially:

1. destination-property ownership,
2. global-versus-target policy behavior,
3. one PP destination per target,
4. legacy `/mcp` coexistence,
5. restart-only discovery, and
6. conflict quarantine behavior?

marianfoo:
continue this in #579

lemaiwo:
Sounds great!

A few remarks:
- for the sysid, sap uses the property sap-sysid for workzone destinations. Ideally we use the same as SAP to be consistent?
- Regarding caching, what about two destinations like in workzone: design time destination for the caching and runtime for the execution with principal propagation (could be in a next step) and we connect both destinations somehow

PR #560, opened 2026-07-09
feat: add authorization trace diagnostics

marianfoo:
## Goal

Add a first-class, read-only `SAPDiagnose action=authorization_trace` workflow for on-prem authorization failures without requiring free SQL.

## What changed

- Added a typed authorization-trace decoder/query module for `SUAUTHVALTRC` plus `TOBJ` field-name mapping.
- Added `authorization_trace` to the SAPDiagnose schema, tool definition, handler, authorization policy, pruning behavior, and frozen tool snapshots.
- Enforced the existing `data` scope and `SAP_ALLOW_DATA_PREVIEW` safety gate; no write, trace-toggle, free-SQL, or cache path was introduced.
- Added graceful absent-table and empty-trace handling while preserving real SAP authorization errors such as HTTP 403.
- Added honest trace-state reporting: ADT cannot read the current kernel/profile value and persisted rows may be historical, so responses report `traceState.status=unknown` instead of guessing.
- Empty responses include RZ11/profile activation guidance for `N`, targeted `F`, broad `Y`, STUSERTRACE filters, multi-instance activation, persistence, and required `S_ADMI_FCD` values.
- Added unit, handler, policy, polluted-input, 403 regression, E2E, documentation, research, and completed implementation-plan coverage.

## Design rationale

ADT does not expose a reusable SU53 runtime endpoint or the STUSERTRACE kernel status methods. ARC-1 already has a bounded, safety-gated table-preview path, so the new action wraps that path with fixed tables, structured filters, client-side ordering, field decoding, and result caps instead of exposing arbitrary SQL.

`SUAUTHVALTRC` rows persist after deactivation. Therefore neither an empty nor a populated query can reliably prove whether `auth/auth_user_trace` is active. The response explicitly exposes that limitation and guides administrators to verify the live value in RZ11.

## Validation

- `npm test`: 138 files, 4,241 tests passed
- `npm run typecheck`: passed
- `npm run lint`: passed
- `npm run validate:policy`: 122 policy entries / 13 schemas passed
- `npm run build`: passed
- `npm run check:sizes`: passed
- `git diff --check`: passed
- HTTP MCP E2E: both authorization-trace tests passed against the actual local server and SAP backend
- Live SAP BASIS 7.58: decoded a failed `S_TCODE` authorization entry and verified the final empty-result warning/setup response
- Live SAP BASIS 8.16: returned a valid empty trace response
- Live parameter proof: `sapcontrol` reported `y` on the 7.58 system and `N` on the 8.16 system; this deliberately remains out of the portable ARC-1 runtime path
- Live safety proof: data-preview-disabled configuration rejected the action before a SAP query


marianfoo:
## Follow-up live SU53/custom-code validation

The table-backed action and an optional custom helper were tested side by side.

- SAP_BASIS 7.58 with `auth/auth_user_trace=y`:
  - Core `authorization_trace` returned the persisted `AUTH_TEST` denial: `S_TCODE`, `TCD=SU01`, RC 12, `LSUSEU11:53`.
  - A temporary `$TMP` helper executed through a real ARC-1 extension returned `TRACE_STATUS=X` and the same live SU53 denial through `SUSR_USER_SU53_READ`, including application-server instance and high-resolution timestamp.
- SAP_BASIS 8.16 with `auth/auth_user_trace=N`:
  - Core action returned the expected empty response with activation guidance.
  - The same helper returned inactive status and no SU53 rows.

The temporary classes were deleted from both systems and their absence was verified.

Conclusion: the core action is useful as a zero-install long-term trace reader. True live trace state and the SU53 shared-memory buffer require custom ABAP. The recommended product shape is an optional on-prem extension plus a separately installed, read-only ICF companion endpoint. A console-class product is not recommended because it cannot accept per-call filters and requires the write/execute safety gates.

Detailed evidence and security tradeoffs are recorded in commit `7dff5237` in the research dossier.

PR #575, opened 2026-07-14
ci: bundle ARC-1 skills inside the .mcpb release artifact

marianfoo:
## What

The `build-mcpb` release job now copies `skills/` into the packed bundle, so the one-click **Claude Desktop (`.mcpb`)** install carries the 22 SAP agent skills alongside the MCP server — parity with the **Claude Code plugin**, which already ships both.

```diff
  cp -r dist mcpb-bundle/dist
+ cp -r skills mcpb-bundle/skills   # ship the ARC-1 agent skills inside the bundle
```

A post-pack guard fails the release if the skills don't survive packing:

```bash
unzip -l "arc-1-${VERSION}.mcpb" | grep -q 'skills/generate-rap-service/SKILL.md' || exit 1
```

## Research findings

- The **MCPB manifest schema has no `skills` field** at any version (checked v0.1–v0.4 in `@anthropic-ai/mcpb@2.1.2`; `latest` is still v0.3). Skills therefore ride as **plain files** inside the bundle, not a manifest-declared capability — this is purely a packaging change, and `mcpb validate` is unaffected.
- `mcpb pack <dir>` zips the **whole directory tree** (respecting `.mcpbignore`), so dropping `skills/` into the staged `mcpb-bundle/` is sufficient to include them.
- Only the 22 product skills under `skills/` are bundled; the repo-dev skills under `.claude/skills/` are excluded (they're not in `mcpb-bundle/`).
- No documented client *auto-activates* skills from an installed `.mcpb` yet, so the `skills/README.md` note keeps `npx skills` / manual copy as the activation path — this PR just makes the artifact self-contained (and future-proofs it).

## Test — full local build of the real bundle

Ran the complete CI flow locally (`npm ci` → `npm run build` → assemble → prune `better-sqlite3` → smoke → `validate` → `pack`):

| Check | Result |
|---|---|
| `mcpb validate` | passes (skills don't affect the manifest) |
| Boot smoke on stdio | passes — `running on stdio` |
| Post-pack skills guard | passes |
| ARC-1 skills in archive | **22/22** under `skills/`, incl. the `references/` subdir |
| Artifact | `arc-1-0.9.27.mcpb`, 9.0 MB packed / 27.4 MB unpacked |

Size impact is negligible — the skills are 568 KB of markdown against a 9 MB bundle.

Incidental observation (pre-existing, not introduced here): `node_modules/dotenv` ships its own `skills/` folder, so two extra `SKILL.md` files have always been present in the bundle. Harmless.

## Files

- `.github/workflows/release.yml` — copy `skills/` into the bundle + post-pack guard
- `docs/publishing-guide.md` — keep the manual repro block + channel table in sync
- `skills/README.md` — MCPB users note now reflects that skills ship in the bundle

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,6 +98,7 @@ jobs:
           cp mcpb-manifest.json mcpb-bundle/manifest.json
           cp icon.png mcpb-bundle/
           cp -r dist mcpb-bundle/dist
+          cp -r skills mcpb-bundle/skills   # ship the ARC-1 agent skills inside the bundle
           cp package.json package-lock.json mcpb-bundle/
           ( cd mcpb-bundle && npm ci --omit=dev --ignore-scripts )
           rm -rf mcpb-bundle/node_modules/better-sqlite3
@@ -131,6 +132,9 @@ jobs:
           VERSION="${VERSION#v}"
           npx --yes "$MCPB" validate mcpb-bundle/manifest.json
           npx --yes "$MCPB" pack mcpb-bundle "arc-1-${VERSION}.mcpb"
+          # Guard: the bundled skills must survive packing (the manifest doesn't list them).
+          unzip -l "arc-1-${VERSION}.mcpb" | grep -q 'skills/generate-rap-service/SKILL.md' \
+            || { echo "::error::Packed MCPB is missing the bundled skills/"; exit 1; }
 
       - name: Attach .mcpb to GitHub Release
         env:
--- a/docs/publishing-guide.md
+++ b/docs/publishing-guide.md
@@ -11,7 +11,7 @@ This document describes how to publish ARC-1 to various MCP server registries, m
 | [Cline Marketplace](#3-cline-marketplace) | Manual | Ready | Cline VS Code extension marketplace |
 | [VS Code / GitHub Copilot](#4-vs-code--github-copilot) | Via MCP Registry | Automatic | VS Code Extensions `@mcp` gallery |
 | [Cursor Marketplace](#5-cursor-marketplace) | Manual | Ready | Cursor IDE built-in marketplace |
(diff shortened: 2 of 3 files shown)


marianfoo:
Closing — **verified empirically that this can't work**, so the bundling is dead weight.

Installed the locally-built `arc-1-0.9.27.mcpb` in Claude Desktop and restarted. The packaging itself worked perfectly — all 22 skills extracted to disk:

```
~/Library/Application Support/Claude/Claude Extensions/local.mcpb.marian-zeis.arc-1/skills/   → 22/22 SKILL.md
```

…but the skills never appear. **Claude Desktop does not read skills from an installed MCPB extension**, and there is no way to make it:

- Desktop's skills actually live in `~/Library/Application Support/Claude/local-agent-mode-sessions/skills-plugin/` — an Anthropic-managed, **account-synced** plugin (`creatorType: "anthropic"`, `enabled: true`), not a local filesystem scan.
- `claude_desktop_config.json` exposes only `mcpServers`, `coworkUserFilesPath`, `preferences` — **no skills key**, so there's no hook to register a folder.
- The MCPB manifest schema has **no `skills` field** at any version (v0.1–v0.4 in `@anthropic-ai/mcpb@2.1.2`; `latest` is still v0.3), so skills can only ride as inert plain files.

The supported channel for Claude Desktop is **Customize → Skills → + → Create skill**, uploading a **ZIP per skill** (`SKILL.md` at the ZIP root) — account-private and server-synced, entirely separate from MCPB.

**Conclusion:** shipping `skills/` inside the `.mcpb` puts bytes in the artifact that nothing loads. The channels that already work are unchanged and remain the answer:
- **Claude Code** → `/plugin marketplace add arc-mcp/arc-1` (MCP server *and* all 22 skills)
- **Any agent** → `npx skills add arc-mcp/arc-1`

Worth revisiting only if an MCPB host ever starts scanning bundles for skills. Recorded here so this isn't re-researched.

PR #571, opened 2026-07-13
feat: add procedural unit source surgery

marianfoo:
## Summary

- add on-prem `SAPWrite(action="edit_unit")` support for named `FORM...ENDFORM` and `MODULE...ENDMODULE` blocks in `PROG` and `INCL`
- support function-group structural includes through `type="INCL", group="<FUGR>"`
- locate and splice complete procedural units with abaplint AST ranges while preserving sibling units and CRLF source
- read the current inactive draft when present, then use the existing package gate, lint/syntax checks, cache invalidation, and lock/modify/unlock write path
- expose the action through authorization policy, on-prem schemas/tool metadata, documentation, and snapshots

## Why

Issue #558 asks for token-efficient edits to classic procedural ABAP without requiring callers to resend an entire report or include. The original proposal assumed ADT object-structure support, but the current implementation uses abaplint because ADT does not expose bounded FORM/MODULE structures.

Event blocks remain intentionally unsupported because abaplint does not expose safe bounded structure nodes for them. `FUNC` remains out of scope because each function module already has its own `/source/main` resource.

Closes #558.

## Final review findings applied

- normalize inactive worklist types so ADT's `PROG/I` correctly matches ARC-1's canonical `INCL`, preserving consecutive draft edits
- reject replacement payloads containing additional procedural blocks; one edit can replace exactly one complete unit

## Validation

- `npm test` — 143 files, 4,332 tests passed
- `npm run typecheck`
- `npm run lint`
- `npm run check:sizes`
- `npm run validate:policy`
- `npm run build`
- live integration test: create → seed → activate → `edit_unit` → reactivate → read back → cleanup for both a PROG FORM and an INCL MODULE

PR #223, opened 2026-05-08
feat!: purge invented ADT slash aliases (issue #218 audit, Plan A)

marianfoo:
## Summary

Implements Plan A from PR #222 — `docs/plans/audit-purge-invented-adt-types.md`. Issue [#218](https://github.com/marianfoo/arc-1/issues/218) follow-up after [@oisee flagged](https://github.com/marianfoo/arc-1/issues/218#issuecomment-4404553535) that `'FUNC/FM': 'FUNC'` was invented.

Five bugs of the same class as `STRU/DS` purged + structural guards added so this class can't recur.

### Removed (invented aliases — pass-through means schema rejects them loudly)
- `'FUNC/FM' → 'FUNC'` — ADT emits `FUGR/FF` for function modules, never `FUNC/FM`
- `'CLAS/LI' → 'CLAS'` — absent from Eclipse apidoc and live ADT responses
- `'VIEW/V'  → 'VIEW'` — ADT emits `VIEW/DV` (verified a4h + npl 2026-05-08)
- `'TRAN/O'  → 'TRAN'` — ADT emits `TRAN/T` (verified a4h + npl 2026-05-08)

### Repointed (real but mis-routed)
- `'FUGR/FF': 'FUGR'` → `'FUGR/FF': 'FUNC'` — `FF` is a function module (LIMU FUNC under FUGR), not the function group. Routing to `FUNC` lets the URL builder append `/fmodules/<fm>`.

### Added (verified-real)
- `'VIEW/DV' → 'VIEW'` (live a4h + npl)
- `'TRAN/T'  → 'TRAN'` (live a4h + npl)
- `case 'VIEW'` in `objectBasePath` returning `/sap/bc/adt/vit/wb/object_type/viewdv/object_name/`. Without this, VIEW reads silently fell through to `/programs/programs/`. Live: `/sap/bc/adt/ddic/views/V_USR_NAME` returns HTTP 500 on both systems; only the VIT URL works.
- `client.getView()` URL changed end-to-end (the URL builder fix alone wasn't enough; the client method also called the broken URL).

### Anti-cargo-cult guards
- `KNOWN_BASE_TYPES` set + exhaustiveness throw inside `objectBasePath` default branch — adding a canonical type to SAPRead/SAPWrite enums without a switch case now fails loudly.
- `SLASH_TYPE_EVIDENCE` map: every key in `SLASH_TYPE_MAP` must point at a `research/abap-types/types/<short>.md` file that exists on disk.
[...]

PR #576, opened 2026-07-14
docs: expose group param for INCL (FUGR structural-include update) & fix for SAPActivate & SAPTransport

frozenraindrop:
## Summary

Fixes a group-routing gap for FUGR structural includes (`type=INCL`, e.g. a FORM-routine or TOP
include inside a function group) across three tools. Rebased onto `main` — this PR previously
also touched `SAPActivate`, but that was independently and functionally-identically fixed by
#571 (merged after this branch was created), so that part has been dropped here. See below.

## What's in this PR now

- **SAPWrite** (`type=INCL` + `group`): the write path already worked (FEAT-18/#505), but the
  tool schema never documented `group` for `type=INCL` — only for `type=FUNC`. A caller had no
  way to discover it, and would fall through to the standalone `/programs/includes/` path,
  hitting the package-resolution fail-closed error. Doc/schema-only fix.
- **SAPTransport** (`check`/`history` + `group`): both actions resolved the object URL via
  `objectUrlForType` directly, with no group-aware routing. For `type=FUNC` this threw loudly
  (the existing `objectBasePath` guard from #223); for `type=INCL` it silently resolved the wrong
  standalone path and returned a misleading "no related or candidate transports (likely $TMP /
  local object)" for an object that is, in fact, package-bound with real transport history. Adds
  a `group` param and routes both actions correctly for FUNC (auto-resolved via search if
  omitted) and INCL.
- A few extra `SAPActivate` regression tests that #571 didn't add: the no-group standalone
  fallback, batch activation with `group`, and the package-allowlist gate — kept since they add
  coverage #571's own test didn't include.

## Dropped (superseded by #571)

`activate.ts` INCL routing, `SAPActivateSchema.group` (top-level + `objects[]`), the `tools.ts`
SAPActivate `group` field, and the duplicate single-object activate test — all already shipped
in #571 with equivalent logic and guards.

## Test plan

- [x] `npm test` — full suite green (only pre-existing, unrelated failures present before and
      after this change)
- [x] `npm run typecheck` / `npm run build` — clean
- [x] `npm run check:sizes` — file-size ratchet + tool-schema wire-byte budget both pass
- [x] Live-verified end-to-end on a real SAP system: `SAPWrite type=INCL group=<FUGR>` write,
      then `SAPActivate type=INCL group=<FUGR>` activation, both succeeding against a real
      function-group structural include

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/scripts/ci/check-file-sizes.mjs
+++ b/scripts/ci/check-file-sizes.mjs
@@ -30,13 +30,13 @@ const BUDGETS = {
   // submodules ride the default src budget; keep this tight so the dispatcher can't reabsorb them.
   'src/handlers/write.ts': 360,
   // tools.ts holds every tool's JSON schema. The #520 description trim (write-mode tools/list
-  // 87→66 KB to clear the Copilot-for-Eclipse gateway limit) shrank it; lowered to match. The
-  // CLIENT-SAFETY size guard is scripts/ci/check-tool-schema-budget.ts — trim there before raising this.
-  // +text-pool SAPWrite actions/description (edit_text_symbols/edit_selection_texts).
-  // +1 line: documented `group` for INCL (FUGR structural-include update) in the SAPWrite schema.
-  // +5 lines: `group` param on SAPActivate for FUGR structural-include activation.
-  // +4 lines: `group` param on SAPTransport check/history for FUGR structural includes.
-  'src/handlers/tools.ts': 1724,
+  // 87→66 KB to clear the Copilot-for-Eclipse gateway limit) shrank it; #571 moved the large
+  // SAPWrite description constants out to tool-descriptions.ts, shrinking it further — lowered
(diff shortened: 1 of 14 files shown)


marianfoo:
Thanks for this PR. 

This branch is based on `b9ba488c`, which predates #571 (merged Jul 13). That PR
independently landed the same SAPActivate fix — functionally identical routing, same
guard — which is why this is conflicting.

**Please rebase onto `main` and drop the superseded parts:** the `activate.ts` INCL
routing, `SAPActivateSchema` `group` (top-level + `objects[]`), the `tools.ts`
SAPActivate `group`, and the single-object activate test. All already there.

Check in detail what actually should be in in this PR

frozenraindrop:
Bless this Claude tools... done, I think.

Issue #250, opened 2026-05-09
Creation of function modules does not work / is not supported.


ghostxwheel:
Hi,

Based on the PR #223 - looks like function objectBasePath in src/handlers/intent.ts return error for type FUNC.
Is it a bug? Or creation of function modules is not supported?

Thanks and BR,
Greg

marianfoo:
seems like it never worked, even before #223 
thank you for your issue, check version 0.9.3 if this is working now for you

ghostxwheel:
@marianfoo Thanks for the quick fix. 
New function module is created successfully, but no parameters can be added via mcp - tool returns that input/output parameters should be set via SE37/Eclipse. 
Can direct parameter addition be added later? Maybe via SAPWrite update?
Thanks again,
Greg

PR #581, opened 2026-07-16
 SAPRead line-range + SAPWrite edit_content, with FUGR-INCL group auto-resolve fixes

frozenraindrop:
## Summary

Three related changes, built and tested together:

1. **`SAPRead` `lineStart`/`lineEnd`** — optional 1-based, inclusive raw line-range read, for
   reading around a known line (e.g. an ATC finding) without paying for the whole object.
   Mutually exclusive with `grep`/`method` for now; `lineEnd` clamps to the last line rather
   than erroring.

2. **`SAPWrite` `action="edit_content"`** — SNOTE-style anchored replace (`oldContent`/
   `newContent`) for `PROG`/`INCL`, for fixes smaller than a whole `FORM`/`MODULE` (`edit_unit`)
   or a full-source `update`. Fail-closed on 0 or >1 matches; a 0-match with a uniquely-present
   `newContent` is treated as an idempotent no-op rather than an error, for retry-safe batch
   remediation loops.

   New primitive `safeUpdateSourceWithTransform` (`src/adt/crud.ts`) fetches the current source
   **inside** the SAP lock, in the same stateful session as the PUT, with an optional conditional
   GET (`If-None-Match`) against a cached `{source, etag}` pair. `edit_unit` is retrofitted onto
   the same primitive, closing a pre-existing TOCTOU gap where it fetched source *before*
   acquiring the lock.

3. **FUGR structural-include `group` auto-resolve** — `edit_content`/`edit_unit`/`update` on a
   FUGR-owned `INCL` only used the package-scoped URL when `group=` was passed explicitly;
   omitting it fell through to a flat URL with no `packageRef`, failing closed against the
   package allowlist. Now auto-resolves the owning group via search when omitted, mirroring
   `FUNC`'s existing `#250` behavior — a null result (a genuinely standalone include) falls
   through to the existing flat path unchanged. The identical gap existed in `SAPActivate`'s
   single and batch `INCL` branches and is fixed the same way. `SAPWrite`'s `group` parameter
   description previously only ever mentioned `FUNC`, which is why this wasn't discoverable
   before.

## Test plan

- [x] Unit tests added/updated across all four areas (line-range read, content-splice,
      content-surgery handler, `crud.ts` primitive, `edit_unit` retrofit, `SAPActivate`
      auto-resolve — single/batch/standalone-fallback cases).
- [x] Full suite green except 13 pre-existing, unrelated Windows-only failures (CRLF/chmod/
      path-separator test-harness issues + one real cross-platform bug in `deny-actions.ts`,
      all predating this branch).
- [x] `npm run typecheck`, `npm run lint`, tool-schema wire-byte budget, and file-size ratchet
      all pass.
- [x] Live-verified end to end against a real on-prem system: `edit_content` write (both with
      explicit `group=` and with it auto-resolved) → inactive draft → `SAPActivate` (group
      required there too, now also auto-resolving) → clean activation → re-read confirms the
      change is genuinely active.


marianfoo:
please create a seperate PR for each of those three changes

marianfoo:
Closing as superseded by the focused PRs #585 and #586. All four commits in this combined proposal have patch-equivalent commits in #586; #585 is the focused home for line-range reads. The edit_unit concurrency fix has since shipped in #831. The remaining anchored editing and function-group resolution work stays open in #586.

PR #363, opened 2026-06-05
fix(handlers): harden tool-arg validation against GPT/OpenAI schema pollution (#360)

marianfoo:
Closes #360.

GPT/OpenAI tool callers over-populate optional schema fields, which made valid `SAPWrite` (and other tool) calls fail — or, worse, silently corrupt data. Every failure was reproduced and verified live on **S/4HANA 2023 (SAP_BASIS 758)** and **NW 7.50**; all are at the Zod validation layer and are fixed there. The fix is purely client-side input normalization — the ADT wire contract (`true`/`false` text for DDIC flags) and `boolToXml`/parser round-trip are already correct (confirmed against three independent ADT clients).

## What was broken (reproduced live)

| Pollution | Before | Now |
|---|---|---|
| `signExists="false"` (stringified bool) | `z.coerce.boolean()` → **`true`** — domain persisted with the **opposite** flag (silent DDIC corruption) | correctly `false` |
| `serviceDefinition: null`, `source: null`, … (strict-mode optional emulation) | `z.X().optional()` **rejects `null`** → hard fail (every unused optional; the Responses API applies strict by default) | stripped → accepted |
| `odataVersion=""` / `category=""` / `visibility=""` | empty enum **rejected** | stripped → accepted |
| `length=""` | `z.coerce.number("")` → silent **`0`** | stripped → omitted |
| `include="definitions"` on a non-CLAS / `delete` / `batch_create` | hard **rejected** | dropped (garbage value on a real CLAS include-write path still errors) |

## The fix

1. **`stripLlmEmptyValues`** — a pre-validation pass that strips `null` and empty/whitespace strings, wired into the single `normalizeTypeArgsForValidation` choke-point in `handleToolCall` (covers standard tools, hyperfocused mode, and the CLI). Preserves real `false`/`0`, sanitizes `objects[]` items, and exempts the two handler-meaningful empty-string fields (`target`, `proposalUserContent`).
2. **`looseOptionalBoolean`** applied to **every** optional boolean (36 sites) — accepts real JSON booleans **and** `"true"/"false"/"1"/"0"/"yes"/"no"`/numeric `0/1`, and rejects non-coercible values instead of mapping them to `true`. Deliberately not `z.stringbool()` (it rejects real booleans, which compliant clients like Claude send).
3. **Inapplicable `include` dropped** in normalization + the `tools.ts` `include` description corrected (it had wrongly advertised `add_method`/`edit_method_signature`/`delete_method`).
[...]

Issue #360, opened 2026-06-05
Make `SAPWrite` robust for GPT/OpenAI clients that overpopulate optional schema fields


pietbrauer:
When `SAPWrite` is exposed to GPT/OpenAI-style tool callers, the model often sends a “polluted” payload: it includes many optional fields from the public schema even when they are irrelevant to the requested action or object type.

This causes valid user intent to fail or be routed incorrectly. The most visible case is non-class writes receiving class-only fields like `include="definitions"`, which can make a metadata/source update fail even though the requested object is not a class.

The tool schema currently exposes a broad union-like surface, but many clients do not reliably omit irrelevant fields. The server should defensively normalize these payloads before validation/dispatch.

## Problem

GPT-style callers tend to include optional fields with empty strings, default booleans, or unrelated enum values. Examples of problematic field pollution:

- Empty optional strings such as `source`, `method`, `group`, `serviceDefinition`, `bindingType`, `transport`, `targetAlias`, etc.
- Class-only fields such as `include`, `method`, `visibility`, and `abstract` on non-`CLAS` writes.
- Metadata fields from one object type leaking into another, for example `odataVersion` or `serviceDefinition` on `DTEL`/`DOMA` operations.
- `batch_create` calls receiving create-like fields at the top level instead of only inside `objects`.
- `delete` calls receiving source/update/create metadata fields even though delete only needs a minimal payload.

This makes otherwise valid tool calls brittle. The model may retry the same invalid call because the error looks like a normal schema validation failure rather than a non-retryable argument-shape problem.

## Affected Scenarios

- `SAPWrite(action="create", type="DTEL")` with unrelated fields from `CLAS`, `SRVB`, RAP, or source-write operations.
- `SAPWrite(action="update", type="DTEL" | "DOMA" | "SRVB" | "MSAG")` with unrelated metadata/source fields.
- `SAPWrite(action="update", type="DDLX" | "DDLS" | "DCLS" | "BDEF" | "SRVD" | "TABL")` with class-only fields like `include`.
- `SAPWrite(action="batch_create")` with polluted top-level fields or polluted metadata entries inside `objects`.
- `SAPWrite(action="delete")` with source, include, activation, RAP, or metadata fields attached.
- Error handling for invalid `SAPWrite` arguments, where the client should be told not to retry unchanged.

## Expected Behavior

The server should accept valid intent even if the client includes irrelevant optional fields, as long as those fields can be safely ignored.

Recommended normalization behavior:

[...]

PR #583, opened 2026-07-17
feat: bound unbounded tool results, reject unknown params, compact JSON

marianfoo:
Every number here is live-measured against a4h (SAP_BASIS 758). Evidence: [`docs/research/2026-07-17-token-optimization.md`](docs/research/2026-07-17-token-optimization.md) · plan: [`docs/plans/2026-07-17-token-optimization-plan.md`](docs/plans/2026-07-17-token-optimization-plan.md).

## The problem

CI guards the tool schema (~20k tokens, one-time) with wire walls, ratchets and a dedicated script. Meanwhile a single `SAPNavigate(references)` returned **968,609 tokens** — 4.8× a 200k context window — with no limit at any layer, and it **silently ignored** a caller's `maxResults`. We were guarding the small number and not the big one.

## Results (live)

| Path | Before | After |
|---|---|---|
| `SAPNavigate(references)` on `CL_ABAP_TYPEDESCR` | 3,003,504 B (~968k tok) | **35,126 B (−98.8%)** |
| `SAPTransport(list)` | 104,383 B (~26k tok) | **14,324 B (−86%)** |
| `SAPContext(impact)` with `maxResults` | ignored | **15,535 → 4,684 B** |
| `SAPRead` with a typo'd param | `success` + 543,318 B silent dump | **rejected, key named** |
| Every JSON result | pretty-printed | **−15% to −37%** |

## What's in it

**1. Bounding.** `references`/`usages`/`impact`/`transport list` are paged and return a complete `total`/`summary` next to a capped page. The total is computed *before* slicing — a capped page must never under-report the blast radius of a change, because "what breaks if I change this?" is a safety question.

SAP can't help: `/usageReferences` declares only `{?uri}` while sibling collections declare `maxResults`. Nine request variants returned **byte-identical** responses, and a bogus URI in the body returned the same bytes — the body is schema-validated but semantically ignored. So paging is necessarily client-side: this bounds what reaches the model, not what SAP computes.

**2. `objectType` now actually filters.** It was dead code — SAP ignores `objectTypeFilter` and nothing filtered client-side, so the parameter silently promised something it never did. Live: `CLAS/OC` narrows 6,644 → 1,724.

**3. Unknown params are rejected** (`.strict()`, 16 top-level schemas) instead of silently dropped, and `additionalProperties: false` is advertised so the contract matches. Worst in *hyperfocused* mode, where the `params` bag has no schema — so the token-saving mode was the one most likely to trigger the worst blowup.

**4. `SAPTransport(list)` summarises by default.** The cost is per-request object lists, not the count: capping 50-of-55 saved 2%, dropping object lists saves 4.7×. `summary=false` restores them.

**5. MCP server `instructions`** — the SDK supports it; we passed none. Claude Code defers MCP tools by default, and server instructions are how a client decides to search for them at all.

**6. Compact JSON** — one `toolJson()` helper, ~90 call sites, via a paren-balanced codemod that only rewrote calls it fully parsed as `(X, null, 2)`.

**7. `format:"structured"` described honestly** — it's a strict superset that is never smaller: larger in **8/8** classes, up to **+1,685%** (`CL_ABAP_MATCHER`: 5k → 91k tokens), yet was sold as the smarter option.

## Two Codex reviews found 12 issues — including my own

The first found 9, the second verified those and found 3 more. Several were the exact bug class this PR exists to kill, reintroduced by it:

- **`SAPContext(impact)` accepted `maxResults` and silently ignored it** — I added the param to a shared schema and wired it only to `usages`.
- **`.strict()` broke our own CLI** — `arc1 source` sent a phantom `flat` arg that no handler read and Zod had been silently swallowing.
- **Descriptions that lied** — `"total" always reports the unfiltered count` (it's filtered); instructions claiming every list action is paged.
- **Padded `objectType` silently dropped interface implementers** — I trimmed at the comparison, but the raw value still reached `augmentInterfaceImplementers` one call earlier.
- **Hyperfocused bypassed the strictness stamp** via an early return.

The pattern: fixing where a bug *shows* rather than where it *comes from*. All fixed at the cause, and the tricky ones are mutation-checked.

## The #360 conflict was assumed, not real

`.strict()` looked risky because of #360 (GPT arg pollution). The real payloads in PR #363 carry **47 known keys and zero unknown keys** — the junk is empty *values*, which `stripLlmEmptyValues` already handles. `.strict()` would have rejected **none** of it. Replaying the actual payload through the built schema is now a regression test. Scoped to top level: `.strict()` verifiably does not cascade, so `batch_create` items stay lenient.

## Deliberately not done

- **No global dispatch truncation cap.** Claude Code doesn't truncate — it *"persist[s] to disk and replace[s] with a file reference"*, which the model can grep. A server-side cap would replace that with a stub. Tested: a blunt cut corrupts JSON and *silently* mutilates ABAP (26 of 212 methods survive, tail reads as ordinary code).
- **No TOON.** The only controlled study ([arXiv 2605.29676](https://arxiv.org/html/2605.29676v1)) shows 2–18% saved for up to **−36pp accuracy**.
- **No tokenizer dep** — `bytes/4` validated to −4%/+2% on ABAP source.
- **CDS `deps` compression deferred** — the finding is true (it emits full DDL) but payloads are 305/959/2,113 tokens: ~1,500 saved vs the 964,000 above, at the highest semantic risk in the set.
- **Field pruning rejected** — dropping `""`/`0`/`false` adds 10% but removes `isResult:false`, which separates a real where-used hit from a structural node. Absent ≠ false.

## Budget

`standard-full-git` 67,951 → **67,959** (wall 68,000); `tools.ts` 1713/1714. **Neither wall raised** — the guard says to trim, and "I need room" is exactly the excuse it exists to refuse. Paid for by cutting prose that restated schema properties or described runtime behavior the error already reports.

## Testing

- **0 failed, 4342 passed** on the tree merged with current `main` (trial-merged locally; clean, no conflicts).
- **e2e run live against a4h**: navigate 9/9, cache 9/9, transport 12/12. CI's `e2e`/`integration` jobs fail at the preflight gate — *"A4H 2025 authenticated ADT discovery returned HTTP 500"*, i.e. the CI target is down and the tests never ran. That's why I ran them locally.
- Bounding, padded-`objectType` and compaction guards are **mutation-checked** — reverting each fix fails them.

## The compaction accuracy gate — run, and it passes

My plan said to gate compaction on an accuracy eval. It's now run rather than argued.

Subject: GPT-5 via `codex exec`. Four real ARC-1 payloads (where-used 25 refs, transport list 12, search 18, DEVC 20), 18 questions with programmatic ground truth. Identical questions, payload and order per arm — **only whitespace differs**. The payload is embedded raw in the prompt, exactly as a tool result reaches a model (deliberately not via the Read tool, which re-formats compact JSON and would confound the arms).

| arm | score |
|---|---|
| pretty | **18/18 (100%)** |
| compact | **18/18 (100%)** |
| | **+0.0pp** |

Nine of the 18 required multi-record traversal — filter-and-count across 25 references, summing `objectCount` across 12 transports, "the LAST object in the array" — i.e. exactly where indentation might plausibly help track structure. All correct in both arms.

**Limitation, stated plainly: this is a ceiling effect.** Both arms scored 100%, so the eval has no headroom — it rules out a *large* effect, not a subtle one. Detecting a few points would need questions hard enough to push the pretty arm below 100%.

Worth recording: the first run scored **0/18 vs 0/18** — a perfect "tie" that was really `codex exec` blocking on a non-TTY stdin and answering nothing. A broken eval and a null result are indistinguishable from the summary line. Method notes are in the plan doc.


marianfoo:
## Second Codex review — verified the first 9, found 3 more

Re-ran an adversarial Codex pass over the fixed tree. It **verified each of the 9 earlier findings** (6 fully fixed, 3 partial) and found 3 new issues. All addressed in a3b41d4b + 357cd4a8, verified live.

### The three partials were all the same mistake: I fixed symptoms, not causes

| # | What was still wrong | Fix |
|---|---|---|
| MED | **Padded `objectType` silently dropped interface implementers.** I trimmed inside `matchesObjectType`, but the *raw* value still reached `augmentInterfaceImplementers` first, whose `/^CLAS/i` check rejects `" CLAS/OC "` — augmentation skipped, implementers gone before the trimmed filter could keep them. Asking for `" CLAS/OC "` on an interface could return "No references found." | Normalized **once at the entry** of `lookupLiveUsages`; only that value is used downstream |
| MED | **Hyperfocused bypassed the `additionalProperties` stamp** — early return, before the loop. The mode most likely to trigger a full-object dump (its `params` bag has no schema) advertised a permissive contract while its runtime rejected unknown keys | Every return path goes through `withStrictSchemas()` |
| LOW | `maxResults` still documented as "usages: max entries" only — never advertised impact's per-bucket semantics or its default of 50 | Documents both |

Plus one new LOW: **empty `references` returned bare text** `"No references found."` while every nonempty result returned the envelope — a consumer parsing it throws on an unused object, and the new server instructions promise a `total`. Now returns `{total:0, shown:0, truncated:false, references:[]}`.

### What Codex confirmed clean

- `boundImpactBuckets` does not mutate caller arrays or corrupt `summary` (slice replaces the reference on a copy; summary aliased but never written)
- The CLI `format:"text"` mapping is behaviourally equivalent to the old ignored `flat` for every type
- `toolJson` has no semantic defect
- Plugin (`Custom_*`) tools stay internally consistent — appended after `getToolDefinitions`, validated by the plugin's own non-strict schema, so no mismatch is introduced

### The pattern worth naming

Every partial was me patching where the bug *showed* rather than where it *came from*. Trimming at the comparison instead of the boundary. Stamping at one return instead of all of them. Both looked fixed and passed my tests; both left the real defect intact one call earlier.

### Gate

`0 failed, 4336 passed` · typecheck 0 · policy ✓ · both ratchets ✓ · schema **67,959 / 68,000**, tools.ts **1713/1714** — **neither wall raised**. Budget paid by collapsing three repeated `Only for action="impact". ` prefixes to the `impact: ` convention already used elsewhere.

The padded-`objectType` fix is mutation-checked: restoring the raw value fails the test.

#584 is merged into this PR, so this is now the single PR against `main` (6 → 8 commits).

PR #584, opened 2026-07-17
refactor: emit compact JSON from tool results

marianfoo:
**Stacked on #583** — base is `feat/token-optimization-bounding`, since this touches the same handler modules. Review/merge #583 first; the base retargets to `main` automatically.

Plan: [`docs/plans/2026-07-17-token-optimization-plan.md`](docs/plans/2026-07-17-token-optimization-plan.md).

## What

Every JSON tool result was pretty-printed with 2-space indent. The model gains nothing from indentation, and it costs 15–37% of the payload.

Measured **live, after #583's bounding** — so these are savings on already-bounded results, not a paper win:

| Result | Before | After |
|---|---|---|
| `SAPNavigate(references)` | 11,856 tok | **8,781 (−26%)** |
| `SAPTransport(list)` | 5,689 tok | **3,581 (−37%)** |
| `SAPRead(DEVC)` | 5,372 tok | **4,404 (−18%)** |
| `SAPSearch` | 2,922 tok | **2,472 (−15%)** |

One `toolJson()` helper in `shared.ts`; 86 call sites across 11 handler modules.

## Why this is safe

It is **whitespace only** — the output is still JSON, which models are trained on. It is explicitly **not** a token-optimized notation like TOON: the only controlled study ([arXiv 2605.29676](https://arxiv.org/html/2605.29676v1)) finds those trade 2–18% savings for up to **−36pp accuracy**, with parallel tool calls collapsing to near zero.

The codemod was **bounded** per AGENTS.md Playbook §6: it only rewrites a `JSON.stringify` call whose argument list it fully parsed and confirmed to be exactly `(X, null, 2)`, tracking string/template literals so parens inside strings cannot fool the paren balancer. Anything unparseable was left alone — nothing was.

Human-facing output is unaffected: audit previews and CLI rendering format at their own layer.

## Field pruning evaluated and rejected

| where-used (100 refs) | tokens |
|---|---|
| pretty (before) | 11,856 |
| **compact (this PR)** | **8,781 (−26%)** |
| + drop empty strings | 7,881 (−34%) |
| + drop `""`/`0`/`false` | 6,649 (−44%) |

Empty-string pruning adds only 7.6% over compaction for a recursive transform on every result. Falsy pruning adds more but removes `isResult:false` — which distinguishes a real where-used hit from a structural tree node — and `line:0`, which means "no line info". **Absent does not mean the same thing as false.** Not worth the semantics for 10%.

## Honest caveat

My plan said to gate this on an accuracy eval. I reasoned about accuracy (still JSON, not a notation change, models parse compact JSON routinely) but did **not** A/B test it against a fixture set. If you want that before merging, say so — this is the one change in the series whose downside is argued rather than measured. The revert unit is this PR alone, which is why it's separate.

## The abaplint failures were never real

I previously reported "2 pre-existing failures" from `@abaplint/core` drift. **That diagnosis was wrong** — there is no drift. The lockfile correctly pins 2.119.65 and matches `package.json`. This worktree simply had no `node_modules` and resolved up to the parent checkout, which sits on an older commit with 2.119.37 installed. After `npm ci`:

**0 failed, 4330 passed.** CI was never affected and no fix was needed, so there is no abaplint change in this PR.

## Testing

- Full suite green: **0 failed, 4330 passed**. Typecheck 0 errors, policy, biome, schema budget, build all pass.
- 6 new tests in `shared.test.ts`, **mutation-checked**: reverting `toolJson` to pretty-print fails 2 of them.
- One test pins that falsy values survive, so a future "prune empty fields" change has to argue with it.

### DIFF
--- a/docs/plans/2026-07-17-token-optimization-plan.md
+++ b/docs/plans/2026-07-17-token-optimization-plan.md
@@ -278,7 +278,15 @@ that proves the conflict is absent; a polluted `batch_create` still passes (item
 
 ---
 
-## P6 — Serialization hygiene (do LAST, only after P1)
+## P6 — Serialization hygiene — SHIPPED (PR 2)
+
+> **Status 2026-07-17: done.** One `toolJson()` helper; 86 sites across 11 handler modules via a
+> bounded codemod. Measured live on already-bounded results: where-used 11,856 → 8,781 (−26%),
+> transport 5,689 → 3,581 (−37%), DEVC 5,372 → 4,404 (−18%), search 2,922 → 2,472 (−15%).
+> **Field pruning rejected on measurement:** dropping empty strings adds only 7.6% over compaction
+> alone, and dropping `""`/`0`/`false` removes `isResult:false` (which separates a real where-used
+> hit from a structural tree node) and `line:0` (meaning "no line info") — absent ≠ false.
+> The "gate on an eval" caveat below stands: accuracy was reasoned about, not A/B tested.
 
 **Verified but small.** 77 `JSON.stringify(x, null, 2)` sites across `src/handlers/`.
 
--- a/src/adt/client.ts
+++ b/src/adt/client.ts
@@ -1497,12 +1497,14 @@ export class AdtClient {
     if (u && !u.includes('@')) this.internalUser = u;
   }
 
-  /** Get system info as structured JSON (user, system details from discovery XML) */
+  /** Get system info as structured JSON (user, system details from discovery XML).
+   *  Compact: this string is returned verbatim to the LLM by SAPRead(type="SYSTEM"). Cannot use
+   *  handlers' toolJson() here — adt/ must not depend on handlers/. */
   async getSystemInfo(): Promise<string> {
     checkOperation(this.safety, OperationType.Read, 'GetSystemInfo');
     const resp = await this.http.get('/sap/bc/adt/core/discovery');
     const info = parseSystemInfo(resp.body, await this.getEffectiveUser());
-    return JSON.stringify(info, null, 2);
+    return JSON.stringify(info);
   }
 
   /** Get installed SAP components */
--- a/src/adt/xml-parser.ts
+++ b/src/adt/xml-parser.ts
@@ -1006,7 +1006,7 @@ export function buildApiReleasePutBody(getXml: string, contract: string, state: 
  * We extract the key fields into a JSON summary:
  * - name, description, OData version (V2/V4), binding type (UI/Web API)
  * - service definition reference, publish status, contract
- */
+ * Compact JSON — returned verbatim to the LLM as SAPRead(type="SRVB").source.*/
 export function parseServiceBinding(xml: string): string {
   const parsed = parseXml(xml);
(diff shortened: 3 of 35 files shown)


marianfoo:
## Codex adversarial review — 9 findings, all addressed

Ran a Codex review over the full `origin/main...HEAD` diff. It found real bugs, including some this branch itself introduced. All fixed in c600247c and verified live against a4h.

| # | Sev | Finding | Fix |
|---|---|---|---|
| 1 | High | `SAPContext(impact)` accepted `maxResults` and **silently ignored it** | Buckets capped (50/bucket, clamped); `summary` stays complete. Live 15,535 → 4,684 B, `downstreamTotal` still 38 |
| 2 | High | `references` shape change broke e2e consumers | navigate e2e + eval scenario updated |
| 3 | High | `transport list` shape/default broke e2e cleanup | cleanup passes `maxResults:1000`; assertions updated |
| 4 | Med | **`.strict()` broke our own CLI** — `arc1 source` / `read --flat` | CLI sends `format:"text"`, not the phantom `flat` |
| 5 | Med | Runtime strict contradicted every advertised schema | `additionalProperties:false` stamped centrally, top-level only |
| 6 | Med | `usageCount` semantics changed; cache e2e asserted `=== usages.length` | Asserts `shown` for page, `usageCount` for total |
| 7 | Low | Server instructions over-claimed "all list actions are paged" | Names the four that are; says the rest aren't |
| 8 | Low | `matchesObjectType` didn't trim → `" CLAS "` matched nothing | Trims; whitespace-only = no filter |
| 9 | Low | Description claimed `total` is the **unfiltered** count | It's the filtered count — description corrected |

Plus Codex's consistency note: migrated the last 3 tool-facing `JSON.stringify()` calls, and compacted two LLM-facing pretty-printers my `src/handlers/`-scoped codemod missed (`getSystemInfo`, `parseServiceBinding`).

### The uncomfortable part

**Findings 1, 7 and 9 are the exact bug class this branch exists to kill** — a param accepted and silently ignored, and descriptions that promise what the code doesn't do. I reintroduced them while fixing them. Finding 1 came from adding `maxResults` to the shared `SAPContext` schema and wiring it only to `usages`.

**Finding 4 is worse**: `.strict()` broke ARC-1's own CLI, because it sent a phantom `flat` arg that no handler ever read and Zod had been silently swallowing for who knows how long. The change that exposed it is the same change that broke it.

**Finding 2/3/6 are on me for only running unit tests.** e2e needs a live server, so I skipped it — and that's precisely where the shape changes bite.

### One reversal

I'd rejected `additionalProperties:false` as "348 bytes of every conversation forever." That's **87 tokens against a 20,130-token schema — 0.4%**. I over-weighted it; Codex was right that advertising a permissive schema while enforcing a strict one is the same dishonesty I spent this branch removing. Added, paid for by trimming prose that restated schema properties.

### Gate

`0 failed, 4334 passed` · typecheck 0 · policy ✓ · both size ratchets ✓ (comments folded, budgets not raised) · schema 67,980 / 68,000 — **wall still not raised**.

PR #589, opened 2026-07-19
fix: answer tools/list immediately instead of blocking on startup discovery

DimiDR:
## Summary
- `tools/list` previously blocked for up to 10s waiting on the startup feature-discovery probe. Some MCP clients (observed: Cline) cancel `tools/list` around ~5s, so the request was cancelled every time and the tool list never populated ("Tools (0)").
- `tools/list` now answers immediately using the currently cached feature state (falling back to the safe on-prem-superset default on the very first call).
- The server declares `capabilities.tools.listChanged` and emits `notifications/tools/list_changed` once discovery finishes, so clients that support live refresh still pick up the discovery-adjusted tool list shortly after.
- Root cause captured via a raw stdio MITM capture between Cline and the server process: `docs/research/2026-07-19-cline-tools-list-timeout.md`.

## Test plan
- [x] `npm test` (added/updated coverage in `tests/unit/server/server.test.ts`)
- [x] `npm run typecheck`

### DIFF
--- a/docs/research/2026-07-19-cline-tools-list-timeout.md
+++ b/docs/research/2026-07-19-cline-tools-list-timeout.md
@@ -1,0 +1,70 @@
+# Cline shows "Tools (0)" for ARC-1 — tools/list timeout race
+
+**Status:** Root cause confirmed via raw stdio capture. Fix: respond to `tools/list` immediately,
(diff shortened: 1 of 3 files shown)


marianfoo:
fixed in #639

Issue #301, opened 2026-05-26
ARC-1 not able to connect to ABAP Cloud from BTP deployment


dingfeldera:
First of all many thanks for this great solution!

I’m trying to setup the following scenario:

MCP Client (Claude Desktop) —> ARC-1 (BTP deployment) —> ABAP Cloud (Trial)

The auth of the client to the ARC-11 BTP deployment was successful (Claude opens to browser etc.) but asking Claude to read a class ends in a timeout because ARC1 is not able to connect to ABAP Cloud (trial). In the BTP Appl Log I can see that there is login attempt which fails:  

_[2026-05-26T08:05:26.556Z] WARN: Could not open browser automatically. Please open this URL manually:_

Opening the URL manually works but the redirect goes to localhost. Any ideas what I’m missing?

Env. Variable `SAP_BTP_SERVICE_KEY` is set for ARC-1 with the service key from my ABAP Cloud Trial instance. SAP system type (`SAP_SYSTEM_TYPE`) is „btp“

Andy

marianfoo:
Actually did not test this setup myself.
ABAP Cloud instance has a destination or how is the connection from arc-1 to ABAP cloud?

dingfeldera:
> Actually did not test this setup myself. ABAP Cloud instance has a destination or how is the connection from arc-1 to ABAP cloud?

yes I've created a destination from service instance (with client credentials as auth method) to the ABAP trial instance and maintained the name of the destination in environment variable `SAP_BTP_DESTINATION`

marianfoo:
@dingfeldera fix merged in main, please check if it works for you

dingfeldera:
sorry for the late reply. 
I've tried whether it is working but unfortunatly now I get Auth error messages (see last log message) while ARC-1 tries to connect the my ABAP Trial instance.

I've created a _OAuth2UserTokenExchange_ destination on space level of destination service. ClientID and Secret is the one from from ABAP Trial service key.

I'm able to connect voa Eclipse ADT to that ABAP environment.

**MCP Inspector** returns folloing error after executin READ tool:

[...]

PR #605, opened 2026-07-21
fix(btp): allow jwt-bearer exchange against ARC-1's XSUAA client

Prolls:
## Context

arc-1 is increasingly consumed by AI orchestration backends (AI assistants, MCP clients, automation pipelines) running on BTP, where end users authenticate via XSUAA SSO. In these setups, the backend needs to propagate the user's identity downstream to arc-1 — not just use a shared service account. The `jwt-bearer` grant type is the standard BTP mechanism for this principal propagation pattern.

Without it, only users whose JWT is directly trusted by arc-1 (e.g. the admin who deployed it) can call the MCP endpoint. All other users are rejected, making multi-user deployments impossible under `SAP_PP_STRICT=true`.

## Problem

When arc-1 is deployed on BTP and consumed by another BTP service, the downstream service needs to exchange the user's JWT for a token scoped to `arc1-mcp!tXXXXXX` before calling arc-1's MCP endpoint. This is required when `SAP_PP_STRICT=true` is set, as arc-1 then enforces that the incoming token's audience matches its own XSUAA app name.

The exchange uses `urn:ietf:params:oauth:grant-type:jwt-bearer` (BTP's standard token exchange grant). Without this grant type declared in `xs-security.json`, the XSUAA instance rejects the exchange with `401 Unauthorized grant type`, leaving the downstream service with no valid token to present — and arc-1 returning `invalid_token: not a valid XSUAA, OIDC, or API key token` on every request.

This is the root cause of the unresolved auth failure reported in #301.

## Fix

Add `urn:ietf:params:oauth:grant-type:jwt-bearer` to the `grant-types` array in `oauth2-configuration`. One line change.

## Test

Verified live on BTP CF (eu10) with `SAP_PP_STRICT=true`: a downstream Node.js backend performs the jwt-bearer mTLS exchange against `certurl`, receives a token scoped to `arc1-mcp!t118776`, and arc-1 accepts it correctly with full principal propagation for all authenticated users — not just the admin.


Prolls:
The CI failures are pre-existing `npm audit` vulnerabilities unrelated to this change (axios, hono, body-parser, fast-uri — also failing on `main`). PR #603 addresses the axios bump. This PR itself touches only `xs-security.json`.

PR #607, opened 2026-07-23
feat: source DCR signing secret from a bound service

MarcusSchoelzel:
Let the DCR signing secret come from a bound Cloud Foundry user-provided service, in addition to the existing flag / env.

On BTP, the DCR signing key falls back to the XSUAA `clientsecret` unless `ARC1_DCR_SIGNING_SECRET` is set. Setting it via `cf set-env` works, but doesn't fit IaC-managed deployments where the secret is provisioned and bound declaratively (Terraform / Vault). This adds a bound-service source so an externally-managed secret survives MTA `cf deploy` without living in the deploy descriptor — no more `invalid_client` re-auth for VS Code / Eclipse users after every redeploy.

## What changed
- When `ARC1_DCR_SIGNING_SECRET` (flag/env) is unset, ARC-1 scans `VCAP_SERVICES` and uses the first bound service exposing a non-empty `signing-secret` credential.
- Resolution precedence: `flag > env > bound service > XSUAA clientsecret`.
- Startup source attribution gains a `service <name>` variant (shown in `config show` and the effective-policy log).

## Behavior
- Backward compatible: with no bound service, behavior is unchanged (`flag > env > XSUAA clientsecret`).
- Empty / whitespace-only `signing-secret` credentials are ignored (treated as not bound).

## Files
- `src/server/config.ts` — `readDcrSigningSecretFromVcap()` + wired into `dcrSigningSecret` resolution.
- `src/server/types.ts` — `ConfigSource` gains a `{ service }` variant.
- `src/cli.ts`, `src/server/effective-policy-log.ts` — render the `service` source.
- `docs_page/xsuaa-setup.md` — bound-service option under "Stable DCR signing key" (CLI + Terraform examples).
- `docs_page/configuration-reference.md` — note the bound-service fallback on the `ARC1_DCR_SIGNING_SECRET` row.
- `tests/unit/server/config.test.ts` — bound-service read, env precedence, empty-credential skip, source attribution.

## Testing
- `npm run typecheck` — pass.
- `npm test` (config suite) — 199/199 pass, including 4 new cases.
- `npm run build` — pass.


marianfoo:
Closing this for now so we do not merge an implementation-first trust-anchor decision. The service-binding direction remains valuable, but we first need to settle deterministic source selection, ownership, deploy/blue-green/delete lifecycle, recovery, rotation, portability, and fail-closed behavior. A replacement documentation PR with the detailed research, the deep review of this PR, and roadmap item SEC-15 will follow immediately.

marianfoo:
Follow-up research and roadmap PR: #649. It preserves the deep review of this branch, evaluates the service-binding proposal alongside the other lifecycle strategies, and tracks the implementation decision as SEC-15.

PR #632, opened 2026-07-27
feat: serve RFC 9728 protected-resource metadata in OIDC mode (#630)

### LINKED ISSUE #630: [Feature]: Expose RFC 9728 protected-resource metadata in OIDC mode (auto-discovery for remote-MCP clients like Claude.ai / Claude Desktop)
## Preflight

- [x] I searched existing issues (open and closed) and this isn't already requested.

## Problem / use case

Running ARC-1 with external OIDC auth (`SAP_OIDC_ISSUER` / `SAP_OIDC_AUDIENCE`, e.g. against Microsoft Entra ID), it exposes **no OAuth discovery metadata**, so only clients that allow a manual authorization-server override can connect:

- `GET /.well-known/oauth-protected-resource` → **404**
- `GET /.well-known/oauth-authorization-server` → **404**
- `401` on `/mcp` → `WWW-Authenticate: Bearer error="invalid_token", error_description="Missing Authorization header"` — **no `resource_metadata` parameter**

As a result, standards-based remote-MCP clients that rely on RFC 9728 auto-discovery — **Claude.ai custom connectors, Claude Desktop custom connectors, and `mcp-remote`** — cannot locate the IdP and fail to authenticate. Only **Claude Code** works, because it lets the user point an explicit `authServerMetadataUrl` at the IdP's `…/v2.0/.well-known/openid-configuration`.

From the source, the discovery routes and the `resourceMetadataUrl` 401 challenge appear to be mounted only inside the `if (config.xsuaaAuth …)` branch in `src/server/http.ts`; the OIDC / API-key branch mounts none of them.

## Proposed solution

Expose the same protected-resource discovery in OIDC mode that already exists for XSUAA, but pointing at the **externally configured** IdP. ARC-1 stays a pure resource server here — no token/authorize endpoints, unlike the XSUAA proxy mode:

1. Serve `GET /.well-known/oauth-protected-resource` (RFC 9728), honoring any path prefix in `ARC1_PUBLIC_URL`:

   ```json
   {
     "resource": "<ARC1_PUBLIC_URL>",
     "authorization_servers": ["<SAP_OIDC_ISSUER>"],
     "scopes_supported": ["<SAP_OIDC_AUDIENCE>/access_as_user"],
     "bearer_methods_supported": ["header"]
   }
   ```

2. On `401`, add `WWW-Authenticate: Bearer resource_metadata="<ARC1_PUBLIC_URL>/.well-known/oauth-protected-resource"`.

Clients then fetch the IdP's own metadata from `SAP_OIDC_ISSUER` (Entra publishes `…/v2.0/.well-known/openid-configuration`) — exactly what already works via the manual `authServerMetadataUrl` override, just discovered automatically instead of requiring per-client configuration.

Notes:
- External IdPs like Entra don't support Dynamic Client Registration, so clients still supply a pre-registered `client_id` (the remote-connector UIs already allow entering Client ID / Secret). This request is the missing **discovery** half only.
- The XSUAA branch is the precedent that the plumbing (`requireBearerAuth({ resourceMetadataUrl })` + the well-known route) already exists — this asks to also expose it for OIDC, with the AS pointing at `SAP_OIDC_ISSUER`.

**Area:** Connection / auth / deployment

**Release-specific?:** No — applies to any OIDC / external-IdP deployment; transport-independent.

## Alternatives / workarounds

- **Claude Code** works today via its manual `authServerMetadataUrl` override.
- **Claude Desktop stopgap:** `npx mcp-remote <url> --header "Authorization:Bearer <token>"` with a static IdP access token (`az account get-access-token --resource api://<clientId>`) — connects, but the token expires (~1h) with no refresh, so it's not a real login.
- **Reverse-proxy shim** in front of ARC-1 serving the RFC 9728 document and injecting the `resource_metadata` challenge on 401 — works, but is extra infrastructure to maintain per deployment.

### Acceptance criteria

- In OIDC mode, `GET /.well-known/oauth-protected-resource` returns valid RFC 9728 metadata listing `SAP_OIDC_ISSUER` as the authorization server.
- A `401` on `/mcp` carries a `resource_metadata` pointer.
- A standards-based client (e.g. a remote custom connector) completes the auth flow against the external IdP with no manual authorization-server URL.

marianfoo:
Closes #630.

## Problem

In OIDC mode (`SAP_OIDC_ISSUER` / `SAP_OIDC_AUDIENCE`) ARC-1 exposed no OAuth discovery and its `401` carried no `resource_metadata` pointer, so only clients with a manual authorization-server override could connect — Claude Code yes, Claude Desktop / Claude.ai custom connectors / `mcp-remote` no. The MCP authorization spec makes RFC 9728 a **MUST** for servers, and `docs_page/oauth-jwt-setup.md:184,282` + `docs_page/auth-test-process.md:137` already described the behaviour, so this was a conformance + docs gap rather than a new capability.

Reproduced on `879376d9` before the change:

| Request | OIDC mode (before) | XSUAA mode (control, same build) |
|---|---|---|
| `GET /.well-known/oauth-protected-resource/mcp` | `404` | `200` + RFC 9728 document |
| `POST /mcp` without a token | `401`, no `resource_metadata` | `401` **with** `resource_metadata` |

Root cause: `resourceMetadataUrl` and every well-known route lived inside `if (config.xsuaaAuth && xsuaaCredentials)` (`src/server/http.ts:364`, `:429`, `:647`–`:658`, `:694`); the API-key/OIDC branch mounted only `/mcp`.

## What this does

Resource-server-only discovery — no token/authorize endpoints, no DCR:

- Serves the document at `/.well-known/oauth-protected-resource/mcp` (RFC 9728 §3.1 path insertion), at the root fallback, and at the `ARC1_PUBLIC_URL` prefix variant, through the MCP SDK's own `metadataHandler` (CORS + `GET`/`OPTIONS` only — same treatment the XSUAA path gets).
- Adds `resource_metadata="…"` to every `401` (MCP and UI guards).
- URLs come from `getAppUrl()` (`ARC1_PUBLIC_URL` / `VCAP_APPLICATION`), never the request `Host` header — a spoofed Host must not be able to aim clients at an attacker's IdP. Covered by a test.
- **`SAP_OIDC_SCOPES`** → `scopes_supported`. Cannot be derived (`access_as_user` is an admin-chosen Entra name; other IdPs differ) and cannot be omitted (MCP clients take the authorize `scope` from the document). Omitted when unset; errors if set without an issuer.
- **`SAP_OIDC_DISCOVERY=false`** → opt-out, see below.
- No `/.well-known/oauth-authorization-server`: ARC-1 is not the AS here. Verified that `https://login.microsoftonline.com/.well-known/oauth-authorization-server/<tenant>/v2.0` 404s while `…/<tenant>/v2.0/.well-known/openid-configuration` returns 200, and the SDK client's discovery chain ends with that append form — publishing the bare issuer is enough.

## Why there is an opt-out

Measured against a live Entra tenant while validating the design: MCP clients send the RFC 8707 `resource` parameter **only when protected-resource metadata exists** (`selectResourceURL` returns `undefined` without it) — which is exactly why Claude Code + Entra works today. Entra rejects that parameter:

```
/authorize …&resource=https://arc1.example.com/mcp    → AADSTS9010010 (resource doesn't match requested scopes)
/authorize …  (no resource, otherwise identical)       → 200 login page
```

(`common` endpoint returns `AADSTS901002: The 'resource' request parameter is not supported`.)

No server-side value fixes this — clients reject a `resource` that isn't same-origin with the server they called, so an `api://<client-id>` value is impossible. `SAP_OIDC_DISCOVERY=false` therefore restores today's exact behaviour for affected tenants. Default stays on: the spec requires the metadata, and IdPs that ignore unknown authorization parameters are the common case.

## Verification

- 7 new tests in `tests/unit/server/http-oidc-metadata.test.ts` (real `startHttpServer` boot + supertest): document at all three paths, 401 pointer, `scopes_supported` present/absent, prefix-aware public URL, spoofed-`Host` immunity, opt-out and api-key-only modes mount nothing. 3 new config tests. Full suite: 4657 passing, typecheck/lint/size-ratchet clean.
- Live against the built `dist/`: document served, `OPTIONS` → `204` + `Access-Control-Allow-Origin: *`, `POST` → `405`, `/.well-known/oauth-authorization-server` still `404`, opt-out yields `404` + no pointer.
- End-to-end through the **real MCP SDK client** functions: `401` → `resource_metadata` → PRM → Entra AS metadata resolved from the issuer → scope taken from `scopes_supported`.

## Known limits (documented, not fixable here)

1. Entra + any SDK-based client needs `SAP_OIDC_DISCOVERY=false` if it hits `AADSTS9010010`.
2. Claude.ai connectors behind Entra have an upstream failure where the authorization code is never exchanged — anthropics/claude-ai-mcp#506.
3. Entra's v2.0 metadata omits `code_challenge_methods_supported` (measured on tenant / `organizations` / `common`); SDK 1.29.0 tolerates that, a spec-strict client would not.

Full evidence, decisions and rejected alternatives: `docs/research/issues/630-oidc-protected-resource-metadata.md`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -167,6 +167,11 @@ SAP_TRANSPORT=stdio              # or: http-streamable
 # SAP_OIDC_ISSUER=https://login.microsoftonline.com/TENANT/v2.0
 # SAP_OIDC_AUDIENCE=arc-1
 # SAP_OIDC_CLOCK_TOLERANCE=0
+# Scopes advertised to MCP clients as scopes_supported (IdP scope names, not ARC-1 scopes)
+# SAP_OIDC_SCOPES=api://CLIENT_ID/access_as_user
+# RFC 9728 auto-discovery (default true). Set false for Entra ID if OAuth fails with AADSTS9010010
+# — Entra rejects the RFC 8707 `resource` parameter clients send once metadata exists.
+# SAP_OIDC_DISCOVERY=true
 
 # ── XSUAA (BTP) ──
 # SAP_XSUAA_AUTH=true
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,11 +4,6 @@ on:
   push:
     branches: ['main']
   workflow_dispatch:
-    inputs:
-      tag:
-        description: 'Image tag (default latest). Use e.g. pr-632 to publish a branch build for testers.'
-        required: false
-        default: 'latest'
 
 jobs:
   # Build Docker images natively per architecture (no QEMU emulation).
@@ -30,15 +25,6 @@ jobs:
             runner: ubuntu-24.04-arm
 
     steps:
-      # `latest` is what every unpinned `docker pull` gets. A dispatch from a
-      # feature branch that forgets the tag input would silently replace it with
-      # an unreleased build, so refuse before anything is pushed.
-      - name: Refuse to publish latest from a non-main ref
-        if: github.ref != 'refs/heads/main' && (inputs.tag == '' || inputs.tag == 'latest')
-        run: |
-          echo "::error::Dispatching from ${{ github.ref }} needs an explicit tag input (e.g. pr-632), not latest."
-          exit 1
-
       - uses: actions/checkout@v7
 
       - name: Set up Docker Buildx
@@ -57,7 +43,7 @@ jobs:
         with:
           images: ghcr.io/arc-mcp/arc-1
           tags: |
-            type=raw,value=${{ inputs.tag || 'latest' }}
+            type=raw,value=latest
 
       - name: Build and push by digest
         id: build
@@ -138,7 +124,7 @@ jobs:
         with:
           images: ghcr.io/arc-mcp/arc-1
           tags: |
-            type=raw,value=${{ inputs.tag || 'latest' }}
+            type=raw,value=latest
 
       - name: Create manifest list and push
         working-directory: ${{ runner.temp }}/digests
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,6 +4,11 @@ on:
   push:
     branches: ['main']
   workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Image tag (default latest). Use e.g. pr-632 to publish a branch build for testers.'
+        required: false
+        default: 'latest'
 
 jobs:
   # Build Docker images natively per architecture (no QEMU emulation).
@@ -25,6 +30,15 @@ jobs:
             runner: ubuntu-24.04-arm
 
     steps:
+      # `latest` is what every unpinned `docker pull` gets. A dispatch from a
+      # feature branch that forgets the tag input would silently replace it with
+      # an unreleased build, so refuse before anything is pushed.
+      - name: Refuse to publish latest from a non-main ref
+        if: github.ref != 'refs/heads/main' && (inputs.tag == '' || inputs.tag == 'latest')
+        run: |
+          echo "::error::Dispatching from ${{ github.ref }} needs an explicit tag input (e.g. pr-632), not latest."
+          exit 1
+
       - uses: actions/checkout@v7
 
       - name: Set up Docker Buildx
@@ -43,7 +57,7 @@ jobs:
         with:
           images: ghcr.io/arc-mcp/arc-1
           tags: |
-            type=raw,value=latest
+            type=raw,value=${{ inputs.tag || 'latest' }}
 
       - name: Build and push by digest
         id: build
@@ -124,7 +138,7 @@ jobs:
         with:
           images: ghcr.io/arc-mcp/arc-1
           tags: |
-            type=raw,value=latest
+            type=raw,value=${{ inputs.tag || 'latest' }}
 
       - name: Create manifest list and push
         working-directory: ${{ runner.temp }}/digests
--- a/docs/research/issues/630-oidc-protected-resource-metadata.md
+++ b/docs/research/issues/630-oidc-protected-resource-metadata.md
@@ -147,7 +147,17 @@ router does not apply.
 2. **Claude.ai custom connectors + Entra** have a separate, upstream-tracked failure:
    [anthropics/claude-ai-mcp#506](https://github.com/anthropics/claude-ai-mcp/issues/506) — discovery
    succeeds, the user authenticates, then the code is never exchanged at `/token`. Not something ARC-1
-   can fix.
+   can fix. **There is a working escape route**, raised by the reporter on PR #632 and verified against
+   the thread: the **token-broker pattern**
+   ([comment](https://github.com/anthropics/claude-ai-mcp/issues/506#issuecomment-4899697146), posted by
+   `@jvitti93` 2026-07-07, confirmed working by `@brinawebb` the same day) — your own minimal
+   OAuth 2.1 AS in front, federating to Entra privately as a confidential client, so the connector never
+   sees Entra. ARC-1 needs no change (it stays a resource server validating the Entra token the broker
+   attaches), but two ARC-1-specific conditions apply and are now in
+   `docs_page/oauth-jwt-setup.md`: the broker owns discovery (so ARC-1 should run with
+   `SAP_OIDC_DISCOVERY=false`), and the broker must federate the **user's** identity — a
+   `client_credentials` shortcut collapses every MCP user into one identity and silently guts per-user
+   scopes, the audit trail, and Destination principal propagation.
 3. Entra's v2.0 metadata omits `code_challenge_methods_supported` (measured on tenant, `organizations`,
    `common`). SDK 1.29.0 tolerates the absence; a client that follows the 2025-11-25 "MUST refuse" rule
    would not.
--- a/docs/research/issues/630-oidc-protected-resource-metadata.md
+++ b/docs/research/issues/630-oidc-protected-resource-metadata.md
@@ -1,0 +1,220 @@
+# Issue #630 — RFC 9728 protected-resource metadata in OIDC mode
+
+**Status:** Confirmed gap, reproduced live 2026-07-27 on HEAD (`879376d9`, v0.9.27); **implemented** the
+same day (branch `feat/oidc-protected-resource-metadata`). Reporter's diagnosis of the code is accurate.
+Additionally: the shipped docs already claimed this worked — a doc/implementation divergence.
+**Issue:** [arc-mcp/arc-1#630](https://github.com/arc-mcp/arc-1/issues/630) — external reporter (`vateseeb`).
+**Classification:** `feature-request` (small, well-scoped) + `docs-bug` + **MCP spec conformance gap**.
+
+## TL;DR
+
+- In OIDC mode ARC-1 served **no** discovery and its `401` carried **no** `resource_metadata` pointer.
+  XSUAA mode has both. Root cause: everything lives inside the `if (config.xsuaaAuth && …)` branch.
+- The MCP authorization spec (2025-06-18 **and** 2025-11-25) says *"MCP servers **MUST** implement OAuth
+  2.0 Protected Resource Metadata (RFC 9728)"*, so this was a conformance gap, not just a convenience ask.
+- **The finding that shaped the design:** the MCP TS SDK sends the RFC 8707 `resource` parameter **only
+  when PRM exists** (`selectResourceURL` returns `undefined` without it) — and **Microsoft Entra ID rejects
+  that parameter outright**. Measured on a live tenant: `AADSTS9010010`; on `common`: `AADSTS901002: The
+  'resource' request parameter is not supported`. The identical request without `resource` returns the
+  login page. So publishing PRM naively would have *broken* the one client combination that works today
+  (Claude Code + Entra) while fixing the others.
+- Shipped: PRM on by default (spec MUST) + `SAP_OIDC_DISCOVERY=false` as a documented, evidence-backed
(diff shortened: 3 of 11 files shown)


vateseeb:
Thanks for turning this around so fast — we filed #630 and run ARC-1 in exactly this setup (OIDC against single-tenant Entra ID, on Azure Container Apps behind an App Gateway path route). A couple of notes from the field:

**Confirmed.** Your `AADSTS9010010` analysis matches ours: with PRM present the SDK client sends the RFC 8707 `resource` parameter and Entra rejects it; without PRM (today) Claude Code works. Docs coverage is great — `.env.example`, the `SAP_OIDC_DISCOVERY` row in the configuration reference, and the "Microsoft Entra ID caveat" section make the footgun + opt-out clear.

**One addition that might help future Entra readers — a path forward, not just the blocker.** The docs frame anthropics/claude-ai-mcp#506 as out-of-scope ("outside ARC-1"). There's a confirmed workaround in that thread that *does* get Claude.ai / Desktop working behind Entra: the **token-broker pattern** — front Entra with your own minimal OAuth 2.1 AS that Claude talks to, federating to Entra privately as a confidential client. Claude only ever sees your compliant AS (static client, no DCR, `/authorize` + `/token` served at the host root), so both the connector code-exchange bug *and* Entra's resource-parameter / DCR limitations stop mattering; your AS then attaches the real Entra token when forwarding to ARC-1 (which stays a pure resource server — no ARC-1 change needed). One user in that thread confirms it works.

Might be worth a one-line pointer in the "Known limits" / Entra caveat so Entra users get a way forward rather than a stop sign. Either way — this is the right conformance fix, and we're keeping it in view for when #506 lands.

marianfoo:
Thanks — that's a useful correction, and you're right that the docs shouldn't stop at "outside ARC-1".

Added in 8772ea99, under the Entra caveat — with two conditions that are specific to running it in front of
ARC-1 and that the upstream thread doesn't cover:

- **Let the broker own discovery.** It's the AS the connector must reach, so it serves the protected-resource
  metadata; run ARC-1 with `SAP_OIDC_DISCOVERY=false` so a passed-through `.well-known` request can't answer
  with a document pointing at Entra instead.
- **Keep the user's identity.** "Attaches the real Entra token" only preserves the user if the broker gets that
  token through a user-facing flow. A `client_credentials` shortcut collapses every MCP user into one identity,
  and ARC-1's per-user scopes, audit trail, and Destination principal propagation all key off the token's user —
  they'd degrade to a single shared user without any error surfacing.

Also noted what it costs: an internet-facing component you build, operate, and secure, holding confidential-client
credentials for your tenant. Worth it if you need Claude.ai connectors today; not the default recommendation.

Since you run this exact topology (single-tenant Entra, Container Apps behind an App Gateway path route) — if you
try the pre-release build, the `ARC1_PUBLIC_URL`-with-base-path leg is the one I'd most like a real-world
confirmation on. It's covered by a test, but yours is the first deployment I know of with a gateway path prefix.

vateseeb:
Nice — the two extra conditions (broker owns discovery + run ARC-1 with `SAP_OIDC_DISCOVERY=false` behind it; preserve user identity via a user-facing flow, not `client_credentials`) are exactly the traps we'd have hit. Good additions.

Happy to give you the real-world confirmation on the base-path leg — we run precisely that topology (single-tenant Entra, Container App behind an App Gateway path route where the gateway strips the `/sap-mcp` prefix before it reaches ARC-1, and `ARC1_PUBLIC_URL=https://<host>/sap-mcp`). That inbound strip is the interesting part: the client/gateway hits `…/sap-mcp/.well-known/oauth-protected-resource/mcp`, but ARC-1 receives it at the root — so the doc has to be reachable there **and** emit the `/sap-mcp`-prefixed public URLs from `ARC1_PUBLIC_URL`.

One blocker on our side: there's no pullable image for this yet — ghcr only has the tagged releases (0.9.20–0.9.27), nothing for `feat/oidc-protected-resource-metadata` / `8772ea99`. Could you push a pre-release or branch image tag to `ghcr.io/arc-mcp/arc-1` (any tag is fine)? We'll mirror it into our ACR and deploy it to our dev environment.

What we'll report back (curl, no OAuth login needed):
- `GET https://<host>/sap-mcp/.well-known/oauth-protected-resource/mcp` (+ the root and prefix variants) → `200`, and whether `resource` / the metadata URLs carry the `/sap-mcp` prefix or the internal stripped path;
- `401` on `…/sap-mcp/mcp` → whether `resource_metadata` points at the correct prefixed URL;
- config used: `SAP_OIDC_DISCOVERY=true`, `SAP_OIDC_SCOPES=api://<client-id>/access_as_user`.

I'm available to run this **this week**, so if you can push a tag in that window we can get you the confirmation quickly. Just say the word with a tag.

marianfoo:
Image is up: `ghcr.io/arc-mcp/arc-1:pr-632` (digest `sha256:85e635a9…`), built from the branch. linux/amd64 only,
and it's a hand-built dev image — unsigned, no provenance, don't promote it past your dev environment.

If you'd rather build it yourself (or need arm64), same thing:

    git clone -b feat/oidc-protected-resource-metadata https://github.com/arc-mcp/arc-1
    cd arc-1 && docker build -t arc-1:pr-632 .

I ran your topology through that exact image before pushing it (`ARC1_PUBLIC_URL=https://sap.example.com/sap-mcp`,
prefix stripped):

    /.well-known/oauth-protected-resource/mcp           -> 200  resource: https://sap.example.com/sap-mcp/mcp
    /.well-known/oauth-protected-resource               -> 200  same doc
    /.well-known/oauth-protected-resource/sap-mcp/mcp   -> 200  same doc
    POST /mcp -> 401  resource_metadata="https://sap.example.com/sap-mcp/.well-known/oauth-protected-resource/mcp"

So the prefixed URLs come from `ARC1_PUBLIC_URL` no matter which form reaches ARC-1, the internal stripped path
never leaks into the document.

One thing about your gateway before you conclude anything. The MCP SDK builds the well-known URL by inserting the
path at the host root, so for `https://<host>/sap-mcp/mcp` it fetches
`https://<host>/.well-known/oauth-protected-resource/sap-mcp/mcp` — that URL does not start with `/sap-mcp`, so
your App Gateway route probably won't forward it and you'd get a gateway 404 that looks like an ARC-1 bug. ARC-1
serves that path fine, it just has to reach it, so add a route for `/.well-known/oauth-protected-resource*` if you
want that fallback to work. Clients that call `/mcp` first and follow `resource_metadata` don't care, that's what
Claude Desktop and mcp-remote do.

For the report: paste the `WWW-Authenticate` header as is rather than "points at the right url", the quoting is
what clients parse.

marianfoo:
@vateseeb any news?

marianfoo:
i´ll prepare v1.0 and i merge so this will be included here, if you need mroe changes create a new issues or PR

vateseeb:
Confirmed on real hardware — thanks for the `pr-632` image (`ghcr.io/arc-mcp/arc-1:pr-632`, digest `sha256:85e635a9…`). Mirrored it into our ACR and ran it in dev behind the App Gateway path route, exactly the topology in question: gateway strips `/sap-mcp` before ARC-1, `ARC1_PUBLIC_URL=https://dev.paip-api.palfinger.com/sap-mcp`, `SAP_OIDC_DISCOVERY=true`, `SAP_OIDC_SCOPES=api://<client-id>/access_as_user`. All requests below go through the gateway from outside.

**`GET /sap-mcp/.well-known/oauth-protected-resource/mcp` → 200**
```json
{"resource":"https://dev.paip-api.palfinger.com/sap-mcp/mcp","authorization_servers":["https://login.microsoftonline.com/<tenant>/v2.0"],"bearer_methods_supported":["header"],"resource_name":"ARC-1 SAP MCP Server","scopes_supported":["api://<client-id>/access_as_user"]}
```

**`GET /sap-mcp/.well-known/oauth-protected-resource` (no `/mcp`) → 200** — identical document.
**`GET /sap-mcp/.well-known/oauth-protected-resource/sap-mcp/mcp` (prefix-variant) → 200** — identical document.

**`POST /sap-mcp/mcp` → 401**, header verbatim:
```
www-authenticate: Bearer error="invalid_token", error_description="Missing Authorization header", resource_metadata="https://dev.paip-api.palfinger.com/sap-mcp/.well-known/oauth-protected-resource/mcp"
```

So it's exactly as you found: `resource` and `resource_metadata` carry the `/sap-mcp` prefix from `ARC1_PUBLIC_URL` regardless of which form reaches ARC-1, and the internal stripped path never leaks into the document. The base-path leg holds up end-to-end through a live App Gateway path prefix.

**On your gateway note — confirmed, and it's on our side, not ARC-1.** The SDK's root-inserted lookup `GET /.well-known/oauth-protected-resource/sap-mcp/mcp` doesn't start with `/sap-mcp`, so our current route doesn't forward it — we get a gateway **502** (not even a 404), which would indeed look like an ARC-1 bug if you didn't know to check the route. ARC-1 serves that path fine once it's reachable (the prefix-variant above proves it). We'll add a `/.well-known/oauth-protected-resource*` route on the gateway as a follow-up so the root-inserted fallback works for connectors that don't hit `/mcp` first. Clients that call `/mcp` and follow `resource_metadata` (Claude Code here) work today.

dev is already reverted to the pinned release. Thanks for building v1.0 with this in — appreciate the fast turnaround.

marianfoo:
Thanks for running this through a real App Gateway and for including the verbatim responses — this is exactly the end-to-end confirmation we needed. It confirms the base-path handling and `WWW-Authenticate` challenge outside the test harness, including the important distinction between an ARC-1 response and a gateway-side `404`/`502` on the host-root well-known route. Much appreciated.

PR #640, opened 2026-07-30
feat: read Launchpad App Descriptor Items (UIAD) and fix the SDO discovery gate

marianfoo:
## Why

SAP deprecated the classic FLP tile/target-mapping model as of S/4HANA 2023. Its successor is the **Launchpad App Descriptor Item** (LADI, `R3TR UIAD`), and LADI-based content is a prerequisite for SAP Build Work Zone **content exposure v2**.

ARC-1 previously exposed only the deprecated side (`SAPManage flp_*` over `PAGE_BUILDER_CUST`) — and mislabelled it as "business catalogs". On ABAP Cloud, SAP documents LADIs as a developer-owned ADT artifact ("create launchpad app descriptor items manually in ABAP development tools for Eclipse"), which puts them squarely in ARC-1's domain.

## What

**1. `SAPRead type=UIAD`** — registers UIAD as a server-driven object. Registration is a registry entry only: the read/write branches are generic and the SAPRead/SAPWrite type tables derive from `SDO_TYPES`.

**2. Fixes a pre-existing gate bug affecting all 7 SDO types.** `supportsServerDrivenObject()` returns `undefined` on a cold discovery map — which the CLI always is — and callers only short-circuited on an exact `false`. Pre-8.16 systems therefore got a raw 404 with a misleading *"verify the name exists"* hint, when the type doesn't exist on that release at all. `ensureServerDrivenSupport()` resolves discovery first, and is applied at **three** entry points — including `SAPActivate`, which was never gated.

**3. Corrects the `flp_*` descriptions** — they operate on FLP *designer* catalogs (business catalogs are `/UI2/FLPCM_CUST`) — and adds a deprecation pointer to UIAD.

## Live-verified on a4h-2025 (SAP_BASIS 816)

| Fact | Evidence |
|---|---|
| Accept is **blues v2** | v1 → 406; discovery `app:accept` confirms |
| `adtcore:type` = `UIAD/TYP` | metadata GET on real instances |
| 1034 UIAD instances | TADIR |
| AFF source is the exposure-v2 payload | `generalInformation{appType,catalogId,transaction}`, `navigation{targetMappingId,semanticObject,action,form factors}`, `tiles[]` |
| **Writes refused on-prem** | create → `400 Editing of LADIs with ALV "Standard" not allowed in workbench tools` — LADI edits need the ABAP Cloud language version |

Because SAP's refusal is precise, there is **no client-side write guard** — it also covers the read-only manifest-generated items. The "verified 201 for all registered types" docstring was amended rather than inherited.

### Gate verification

| Case | Result |
|---|---|
| `SAPRead type=UIAD` on 816 | metadata + AFF JSON ✅ |
| `SAPRead type=UIAD\|DESD` on 758 | release message, not a raw 404 ✅ |
| `SAPActivate type=DESD` on 758 | release message (newly gated) ✅ |
| **`SAPRead type=EVTB` on 758** | still reads — no false-negative gating ✅ |

New integration test passes live on 816 and skips cleanly on 758.

## Design notes

- `safety` is a **required** param on the gate: `fetchDiscoveryDocument` issues an unguarded GET and this now runs inside a tool call.
- The fetched map is **deliberately not stored** — `server.ts` re-injects the cached map per call, and writing to the shared client would leak one user's capability view under the non-strict PP fallback.
- An empty map (discovery unreachable) resolves to *proceed*, so a 403'd `/sap/bc/adt/discovery` can't block SDO reads. The gate improves error text; `checkOperation` and the package allowlist remain the real controls.
- Schema-budget `descriptionTokenEstimate` raised +100 on two scenarios, documented in-place, after twice compressing the prose.

## Rejected (researched)

| Rejected | Reason |
|---|---|
| UIAC (technical catalog) read | No ADT endpoint on 816 — discovery lists only `fiori/{uiad,uipgtyp,uisttop}` |
| UIPG / UIST registration | 1 instance each — speculative |
| A LADI migration audit tool/skill | SAP ships `/UI2/MGR_TC` (per-catalog migration status) and `/UI2/FLPCA` (inventory, spreadsheet export) |
| Reading `SUI_TM_MM_*` tables | Undocumented internal tables; the supported ADT read supersedes them |

Plan with full findings: `docs/plans/ladi-uiad-read-and-sdo-gate.md`.

## Follow-up (pre-existing, not in scope)

Batch `SAPActivate` (the `objects` array) has no SDO branch, so an SDO type falls through to `objectBasePath` and builds a programs URL. Filed separately.

## Test plan

- `npm test` — 3911 passed
- `typecheck`, `lint`, `validate:policy`, `build`, `check:sizes` — all green
- Integration `-t "816 SDO read"` — 4 passed on 816; UIAD/DESD skip on 758, EVTB passes


marianfoo:
Superseded by a branch built on current `main`. This branch was accidentally cut from `codex/readonly-ui-console`, so it carried 13 unrelated commits — and `main` had since refactored the SDO registry (`metadataContentType` + `discoveryMarker`, plus DSFD/DTDC), which this work needed to be rewritten against. Replacement PR follows.

PR #641, opened 2026-07-30
feat: propagate W3C trace context to SAP and record calling-agent identity

marianfoo:
Closes two named requirements from SAP's Architecture Center guidance for [Third-Party MCP Access to SAP Solutions](https://architecture.learning.sap.com/docs/ref-arch/137800), and records where ARC-1 stands on the other sixteen.

## Why

That page is a permission-with-conditions: customers may use third-party MCP servers against SAP APIs *provided* the API Policy is honoured, auth is enforced on every connection, rate limits are respected, and its hardening is in place. ARC-1 sits in both patterns it names (self-hosted npm/Docker, and the BTP CF deployment), so the whole customer-responsibility clause lands on us.

Evaluating all 18 points found most already met — several above the bar — and exactly two genuine gaps.

## What changed

**W3C trace context now passes through to SAP.** A valid inbound `traceparent` (and `tracestate`) is forwarded verbatim on every outbound SAP call, so one trace spans agent → ARC-1 → SAP.

ARC-1 runs no tracer, so per the [W3C spec](https://www.w3.org/TR/trace-context/) it behaves as a **non-participating pass-through**: it never rewrites `parent-id`, and it **never originates a trace** when the client sent none — `requestId` already correlates ARC-1's own logs, and minting trace-ids no collector knows about is noise. Malformed input is dropped, not repaired.

Validation rejects the spec's forbidden forms (version `ff`, all-zero trace-id/parent-id, uppercase hex), holds version `00` to exactly 55 characters while letting a higher version append fields we forward blind, and restricts the charset so CR/LF header injection is structurally impossible. Injection sits in `doFetch` — the single outbound choke point, which also covers the Cloud Connector proxy branch.

**Audit events now record the calling agent.** `clientAgent` sits beside the registered `clientId` on every audit event and in the BTP Audit Log, including the pre-dispatch denial events (`safety_blocked`, `auth_scope_denied`, `mcp_rate_limited`) where "which agent did this" matters most.

Resolution is MCP handshake `clientInfo` → HTTP `User-Agent` → absent. The fallback is load-bearing: the HTTP transport is stateless, so each POST builds a fresh `Server` that never saw `initialize` and `getClientVersion()` is always `undefined` there — it is exact only on stdio. `clientAgent` is caller-controlled, so it is sanitized, truncated, **audit-only, never sent to SAP, and never an authorization input**.

> **This does not close SAP's requirement, and the docs say so.** SAP asks that the *exchanged token* carry agent identity. ABAP has no claim slot for it, so SAP-side records (SM20, transport owner, `adtcore:changedBy`) still attribute the human. This closes ARC-1's half only.

Both behaviors are unconditional — no new configuration.

## Docs

- `docs/research/2026-07-30-sap-architecture-center-mcp-alignment.md` — the point-by-point evaluation, including the seven things deliberately **not** built (circuit breaker, SAP Credential Store, saga/compensation engine, SAP Passport, a kill-switch env var…) with the reasoning for each, so they don't come back as review comments.
- `docs/security-model.md` — OWASP MCP Top 10 → invariant mapping (snapshot-dated; it's a living document), plus an explicit note that the Cloud Connector `SAP-Connectivity-Authentication` flow hands the user JWT to the *connectivity proxy that exchanges it*, not to a downstream API. It reads like the anti-pattern SAP names and isn't one.
- `docs_page/deployment-best-practices.md` — what actually breaks at `instances > 1` (rate limits become N×, caches go cold, ADR-0007 shared-Basic requires exactly one), and what to alert on.

## Two things left open on purpose

1. **`ARC1_RATE_LIMIT` defaults to 0 (off).** Rate limiting is the one place the SAP page says "must not exceed", and our default is no per-user limit at all. Flipping it is a behavior change — wanted a decision, not a surprise.
2. **The `500 "database connection is not open"` retry fires for all methods.** The 503 retry above it is correctly justified (ICM rejects pre-dispatch, nothing executed); this one isn't proven the same way. Low severity, but "almost certainly did nothing" is the wrong standard for a POST.

## Testing

Full gate green: 165 test files, 4800+ tests, typecheck, lint, `validate:policy`, `check:sizes`, build.

New coverage: `trace-context.test.ts` (spec conformance incl. injection attempts and every forbidden form), `http-trace-context.test.ts` (end-to-end through the real Express handler — proves the `AsyncLocalStorage` context survives into the SDK request handler, the riskiest assumption here), plus extensions to the ADT HTTP, dispatch, logger, context, and BTP-sink suites.

Reviewed independently by Codex, which found four low-severity issues — two W3C conformance gaps (future-version fields, `tracestate` HTAB) and two incomplete-attribution paths — all fixed in 660e562d. It found no security defects and no regressions.


marianfoo:
## Added: SAP API Policy & architecture alignment page

`docs_page/sap-api-policy-and-architecture.md`, under **Reference → Configuration Precedence** in the nav.

The page separates two questions that keep getting conflated:

| | Who answers it | Where ARC-1 stands |
|---|---|---|
| **Is the architecture sound?** | Engineering — measurable | Strong (the 18-point dossier in this PR) |
| **Is the use permitted?** | **SAP** — account team / legal | **Open** |

Grounded in the primary text of **SAP API Policy v.4.2026a**, not secondary reporting. The clauses that actually bite:

- **§1.1/§1.2 Published APIs.** ADT REST (`/sap/bc/adt/*`) is not on the SAP Business Accelerator Hub, and the policy puts the verification duty on the *customer*. This is the central open question — for ARC-1 and equally for `abap-adt-api`, abapGit and abaplint. The on-premise carve-out covers *customer-built* ABAP interfaces, not ADT.
- **§2.2.2** restricts agentic/generative AI that plans and executes sequences of API calls, except through SAP-endorsed architectures. An LLM-driven MCP server is exactly that pattern; whether the Architecture Center third-party MCP guidance is the applicable pathway is a question for SAP.
- **§3** prohibits circumventing API controls via intermediaries — answered head-on: ARC-1 only *adds* controls, principal propagation means no impersonation, and the shared-identity modes are flagged rather than hidden.
- **§2.1/§2.2.1** rate limits, with the honest note that `ARC1_RATE_LIMIT` ships off.

**On the MCP Gateway** — worth flagging, because it corrects a common reading. The Gateway creates MCP servers from three sources: API artifacts on Integration Cell, HTTP endpoints **with an OpenAPI spec**, and RFC backends. All three are *API-to-MCP wrapping*. There is no documented path to register an existing **external MCP server** as an upstream, so fronting ARC-1 with it is not possible today, despite marketing copy describing the Gateway as aggregating "external MCP servers". When that federation lands it closes several gaps at once — central rate limits, SAP-managed spec upgrades, and an SAP-operated component in the path, which materially changes the §2.2.2 conversation.

**Recommendation as stated on the page:** dev/test at your own risk; ask your SAP contact before production — with four specific questions to put to them, because vague questions get vague answers.

Also renames the `instances > 1` heading added earlier in this PR: GitHub and mkdocs slugify it to *different* anchors, so no single link worked in both.

Verified with `mkdocs build --strict` (clean) plus the full test/typecheck/lint/size gate.

PR #643, opened 2026-07-30
fix: route server-driven types in batch SAPActivate instead of the program path

marianfoo:
> **Stacked on #642** — base is `feat/ladi-uiad-read-sdo-gate-v2` because this uses `ensureServerDrivenSupport`, which that PR introduces. Rebase onto `main` once #642 merges.

## The bug

Single-object `SAPActivate` routes server-driven objects (SDO) via the registry href. The batch `objects` resolver had no such branch, so SDO types fell through to `objectUrlForType` → `objectBasePath`.

SDO codes aren't in `KNOWN_BASE_TYPES` and carry no slash, so **neither guard fired** — not the exhaustiveness guard, not the slash-form guard. The call landed on the default arm, which maps unknown non-slash types to the **program path**.

Confirmed live on a4h-2025 (816), by reverting the fix and re-running:

| | `SAPActivate objects=[{type: EVTB, name: S_BUSINESSPARTNER_CHANGE}]` |
|---|---|
| **Before** | `404 at /sap/bc/adt/programs/programs/S_BUSINESSPARTNER_CHANGE` |
| **After** | `Successfully activated 1 objects: S_BUSINESSPARTNER_CHANGE.` |

Affects all registered SDO types: `DESD`, `DTSC`, `CSNM`, `EVTB`, `EVTO`, `COTA`, `DSFD`, `DTDC`, `UIAD`.

## The fix — route, don't reject

The stale comment claimed SDO was *deliberately* left out because batch is "RAP-stack-oriented". But `EVTB` and `EVTO` are RAP **event** objects — they belong in exactly those batches. So routing is the fix that matches the path's own rationale, and it keeps batch consistent with the single-object branch. Comment updated.

Availability is gated **once per distinct type**, before the per-object resolver runs — `ensureServerDrivenSupport` may fetch discovery, and gating inside the resolver would repeat that fetch for every object in the batch.

## Verification

Three unit tests were written **first and observed failing** (the routing test asserted the program URL was present in the activation body):

- batch routes an SDO type through the registry href, never the program path
- batch gates an SDO type on discovery like the single-object path
- batch still routes non-SDO types normally (control — guards against over-broad matching)

Live:

| Case | Result |
|---|---|
| 816 — batch `[EVTB]` | activates ✅ |
| 816 — mixed batch `[EVTB, DESD]` | both activate ✅ |
| 758 — batch `[DESD]` / `[UIAD]` | release message, no program URL ✅ |
| 758 — batch `[EVTB]` (ships on 758) | still activates ✅ |

Gates: 4774 unit tests, typecheck, lint, `validate:policy`, build, size budgets — all green.

## Note

`tests/unit/handlers/write-create-batch.test.ts` went 17 lines over the 3000-line test default, so its budget is raised to 3050 with an in-place comment. Splitting a 3000-line suite felt disproportionate to a three-test regression addition — happy to split instead if you'd prefer.

### DIFF
--- a/scripts/ci/check-file-sizes.mjs
+++ b/scripts/ci/check-file-sizes.mjs
@@ -48,6 +48,10 @@ const BUDGETS = {
   // the default. Keep tight headroom.
   // + getFunctionModuleProperties and the getFunctionGroup pre-7.52 objectstructure fallback.
   'src/adt/client.ts': 1730,
+  // +regression coverage for batch SAPActivate routing server-driven types via the registry href
+  // (the objectBasePath default arm silently built a program URL). Small, deliberate bump over the
+  // 3000 test default; split the SAPWrite create/batch suites if it needs raising again.
+  'tests/unit/handlers/write-create-batch.test.ts': 3050,
 };
 
 const DEFAULT_SRC = 1500;
--- a/src/handlers/activate.ts
+++ b/src/handlers/activate.ts
@@ -225,6 +225,20 @@ export async function handleSAPActivate(
 
   if (args.objects && Array.isArray(args.objects)) {
     const rawObjects = args.objects as Array<Record<string, unknown>>;
+    // Server-driven types need the registry href — objectBasePath(<sdo>) has no case and its
+    // default arm maps unknown non-slash types to the PROGRAM path, so an ungated batch entry
+    // silently activated /sap/bc/adt/programs/programs/<name>. Gate once per DISTINCT type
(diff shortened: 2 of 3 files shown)


marianfoo:
Folded into #642 rather than merged to `main` — the commit `d2df48a2` was fast-forwarded onto the base branch `feat/ladi-uiad-read-sdo-gate-v2`, which GitHub auto-detected as a merge. The change ships as part of #642; nothing landed on `main` from here.

PR #647, opened 2026-07-31
fix: surface class- and program-level AUnit alerts instead of returning []

francocapra:
## Problem

`parseUnitTestResults` only reads alerts that hang off a `<testMethod>` node. Any alert
that lives above that level is dropped, and the function returns an empty array with no
error — so the caller reports "no tests" and the reason the run failed is lost entirely.

Two response shapes hit this:

1. **No `<testClass>` at all.** The run aborted before test discovery (for example an
   object that cannot be tested). The alert explaining why sits at run level.
2. **A `<testClass>` with alerts but no `<testMethod>` children.** Typically a
   `class_setup` failure, which aborts the class before any method executes.

I ran into this against a 7.58 system while chasing a different problem: a batch
activation left a stale negative discovery cache, AUnit returned shape 1, and the tool
reported an empty run. The alert that would have pointed straight at the cause was parsed
and thrown away.

## Change

Both shapes now emit synthetic `failed` rows carrying the alert title:

- run-level alerts become `testClass: '(run)'` / `testMethod: '(alert)'`
- a class with no methods becomes `testMethod: '(class-level alert)'`, with an explicit
  message when there is no alert to report either

A response with neither test classes nor alerts still returns `[]`, so the existing
"no tests ran" path is unchanged.

The title extraction (which handles both a plain string and a `#text` object, depending on
how `fast-xml-parser` renders the node) is factored into an `alertTitle()` helper and the
existing per-method path now reuses it instead of duplicating the branch.

## Two deliberate choices worth flagging

- **Orphan alerts are only collected when there are zero test classes.** If a response
  carried both test classes and unattached alerts, the latter are still ignored. That
  avoids reporting the same alert twice when it is already surfaced under its method; a
  broader sweep would need dedup rules I could not derive from observed responses.
- **`program` is empty on run-level rows.** There is no URI at that level to parse it
  from, and inventing one from the requested object felt worse than leaving it blank.

Happy to change either if you would rather they behaved differently.

## Tests

Three cases added to `tests/unit/adt/devtools.test.ts`, covering run-level alerts, a
class-level alert, and a class with neither methods nor alerts. Full file passes (157
tests), along with `typecheck`, Biome, and the file-size/schema ratchets.

### DIFF
--- a/src/adt/devtools.ts
+++ b/src/adt/devtools.ts
@@ -1083,10 +1083,39 @@ function parseSyntaxCheckResult(xml: string): SyntaxCheckResult {
   };
 }
 
+/** Extract the human-readable title from an AUnit alert node (string or #text object). */
+function alertTitle(alert: Record<string, unknown>): string | undefined {
+  const titleVal = alert.title;
+  if (titleVal == null) return undefined;
+  if (typeof titleVal === 'string') return titleVal;
+  if (typeof titleVal === 'object' && !Array.isArray(titleVal)) {
+    return String((titleVal as Record<string, unknown>)['#text'] ?? '');
+  }
+  return String(titleVal);
+}
+
 function parseUnitTestResults(xml: string): UnitTestResult[] {
   const results: UnitTestResult[] = [];
   const parsed = parseXml(xml);
   const testClasses = findDeepNodes(parsed, 'testClass');
+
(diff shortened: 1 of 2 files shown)


marianfoo:
will continue in #648

PR #656, opened 2026-08-03
docs: document OIDC gateway well-known routing

marianfoo:
## Goal

Document the reverse-proxy routing requirement confirmed during the real App Gateway validation of #632.

## What changed

- Explain that MCP SDK clients may insert the RFC 9728 protected-resource metadata path at the public host root, before an `ARC1_PUBLIC_URL` base path.
- Tell base-path deployments to forward `/.well-known/oauth-protected-resource*` to ARC-1.
- Clarify that a gateway-side `404` or `502` occurs before ARC-1 can serve its prefix-aware metadata.

## Impact

Operators using a gateway base path can configure discovery routing correctly and avoid mistaking an upstream routing failure for an ARC-1 failure.

## Verification

- Cross-checked the documented example against `mountOidcResourceMetadata` and `http-oidc-metadata.test.ts`.
- `npm run docs:build`
- `git diff --check`


marianfoo:
related to #632

PR #579, opened 2026-07-15
feat: add experimental destination-discovered multi-target endpoints

marianfoo:
## Goal

Add an experimental, default-off SAP BTP Cloud Foundry mode that serves many SAP system/client targets from one ARC-1 deployment while keeping every multi-target route mutation-free. This supersedes the unreleased public contract from #543 while retaining Wouter's contribution and branch history.

The design was shaped through #577 and live BTP/SAP testing. The branch also turns the BTP documentation into an administrator-oriented deployment and operations journey for both single-target and multi-target installations.

## Runtime contract

- enable with `ARC1_MULTI_TARGET_ENDPOINTS=true`; an explicitly configured single-target `/mcp` can coexist
- discover BTP **subaccount** destinations marked with exact `arc1.enabled=true` at startup
- expose pinned `/<SYSTEM-OR-ALIAS>/<CLIENT>/mcp` routes and aggregate `/multi/mcp`; no discovered target owns bare `/mcp`
- support strict `PrincipalPropagation` per target and an explicit, default-off shared `BasicAuthentication` exception
- require XSUAA on every multi-target route and authenticate before revealing target membership
- keep multi-target mutation-free regardless of single-target ceilings or Admin scope
- expose six SAP-contacting tools (`SAPRead`, `SAPSearch`, `SAPQuery`, `SAPNavigate`, `SAPDiagnose`, `SAPContext`) plus aggregate-only `SAPTargets`
- require an explicit aggregate `target` on every SAP-contacting call; there is no default or remembered target
- make `SAPTargets` reader-friendly and role-sensitive: compact active targets for readers, secret-safe registry diagnostics for Admin
- keep data and SQL behind the intersection of application ceiling, destination opt-in, XSUAA scope, and SAP authorization
- force cache `none`, standard tool mode, UI/plugins off, and all mutation ceilings off for multi-target v1
- keep zero targets healthy; quarantine invalid config, unsupported auth, duplicates, route collisions, destination shadows, and registries over 256 enabled candidates
- use an immutable startup snapshot; non-secret destination changes require restart, not an MTAR rebuild
- resolve PP and shared Basic request identity without using one as fallback for the other
- share one per-user quota, one per-IP bucket, and one SAP concurrency semaphore per ARC-1 process

[...]

PR #658, opened 2026-08-03
chore: drop the release-as 1.0.0 pin now that 1.0.0 has shipped

marianfoo:
## The problem

`release-please-config.json` still carries `"release-as": "1.0.0"`. That value is authoritative — it overrides commit footers and the computed bump — so **every release PR from here on would be pinned to 1.0.0 and no new version could be cut.**

[7b6fc90b](https://github.com/arc-mcp/arc-1/commit/7b6fc90b27412cf73613dc6eb1aece0cdbcce26d) added it deliberately, to force the pending release past two competing `Release-As` footers in the unreleased range (`1.0.0` on #604, `0.10.0` on #579), and said so in the commit body:

> MUST remove release-as from this config immediately after 1.0.0 is released, otherwise every subsequent release PR is pinned to 1.0.0.

## Verified before removing

- `v1.0.0` is tagged (`cdb86bc0`, 2026-07-31) and `npm view arc-1 version` returns `1.0.0` — the release it was pinning is done
- `.release-please-manifest.json` is `1.0.0`, so the next bump computes from the right base
- The unreleased range `v1.0.0..main` (7 commits: 1 `docs:`, 6 `chore(deps):`) contains **no `Release-As:` and no `BREAKING CHANGE:` footer** — removing the pin cannot resurrect the old `Release-As: 0.10.0` regression. Those 7 commit types are all `hidden` in `changelog-sections`, so the correct outcome after this merges is simply *no release* until a `feat:`/`fix:` lands.

## Other stale pre-1.0 settings

None in the config — `bump-minor-pre-major` and `bump-patch-for-minor-pre-major` were already removed in that same commit, and `SECURITY.md` already lists `1.x` as supported.

The comment on the release-please action step still described those flags as active (`"which sets bump-minor-pre-major: true … while the package is pre-1.0.0"`), which is now false and misleading in the file that governs releasing. Corrected to state what actually applies from 1.0: `feat`→minor, `fix`→patch, breaking→major.

This PR is `chore:` and touches no runtime code, so it does not itself trigger a release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,9 +17,9 @@ jobs:
     steps:
       - uses: googleapis/release-please-action@45996ed1f6d02564a971a2fa1b5860e934307cf7  # v5
         id: release
-        # No release-type here — the action reads release-please-config.json,
-        # which sets bump-minor-pre-major: true so breaking changes only bump
-        # the minor version while the package is pre-1.0.0.
+        # No release-type here — the action reads release-please-config.json.
+        # From 1.0 the pre-major bump flags are gone, so semver applies normally:
+        # feat→minor, fix→patch, breaking→major.
 
   # npm publishes independently of the Docker CVE gate — a HIGH/CRITICAL finding
   # blocks the Docker image but not the npm package. (The scheduled security-scan.yml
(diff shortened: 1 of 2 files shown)


marianfoo:
Folded into #657 — the pin removal now rides there as commit `af043d5f`, alongside the release-notes work it was found in.

PR #671, opened 2026-08-03
feat: add SAPTransport action="diff" for transport review

marianfoo:
Answers **"what did this transport change?"** — per object, the revision written under the transport vs. the one immediately before it, as unified-diff hunks. Previously only object *names* were reachable (`action="get"`); producing a review meant hand-picking version pairs per object.

Modelled on SAP's own ADT MCP tool (`com.sap.adt.tm.backend.diff.TransportDiffService`, ADT 3.60.1, decompiled from `sapse.adt-vscode` 1.1.1), with corrections each proven necessary against live feeds on a4h (SAP_BASIS 758).

## The root bug

`parseRevisionFeed` matched atom rel `…/relations/transport**s**`. The real feed uses `…/relations/transport/**request**` and carries the id in `adtcore:name` — the link `title` is the transport *description*:

```xml
<atom:link adtcore:name="A4HK906289" rel="http://www.sap.com/adt/relations/transport/request"
           href="/sap/bc/adt/cts/transportrequests/A4HK906289" title="ARC-1 demo: …"/>
```

So `RevisionInfo.transport` was **always** absent on 7.58. The existing test passed only because its hand-written fixture encoded the same wrong assumption — the new tests run against feeds captured from the live system. Legacy rel and an href-tail fallback are kept for other releases.

## Corrections over SAP's algorithm

| Correction | Why it was needed |
|---|---|
| Match request **∪ tasks** | Version records reference either — one demo object names the request, another names a *task* of the request under review. SAP matches one exact `transportNumber` and silently falls back. |
| Sort undated revisions as **oldest** | `CERTRULE_DYNP` version `00001` has no `<atom:updated>` at all. Falling back to the version number let it outrank a newer dated revision *and* made the comparator non-transitive. |
| Skip **every** same-transport sibling walking back | SAP skips at most one active entry, so two saves under one transport diff only the last save. |
| Predecessor needs **no** CTS id | Released revisions with a blank transport exist (reproduced independently of the PoC's `T4DK9A21AJ` finding). |
| Break timestamp ties on version number | The feed is returned in neither version nor date order — a4h returns `00002, 00000, 00001`. |

## Review quality

- **LIMU rollup** — `METH`/`CINC`/`CPUB`… roll up to the owning class, handling both CTS name forms (`ZCL_X            SUBTRACT`, `ZCL_X============CCIMP`). A transport lists methods; a review needs classes.
- **Classes diff across all five includes**, so a change in CCIMP or the test class is not silently missed.
- **`selectionMethod` + `baselineStatus` per part** — exact evidence vs. fallback guess. Creation is asserted only when the revision was matched to the transport *and* has no predecessor; a missing baseline is `baseline-ambiguous`, never "newly created". SAP's output schema reports neither.
- **Untouched class includes are listed but not diffed**, so one real change is not buried under four blocks of SAP boilerplate.
- **Non-source objects are reported as inventory**, not dropped or errored.

Unreleased transports need no special case: while objects are locked the active revision carries the transport link, so the same path yields last-released → active. (The planned `active↔inactive` branch was written, then deleted once live testing showed it redundant.)

## Safety

`read` scope, `Read` opType, `transport` feature gate — identical to `SAPTransport.get`. No new ADT endpoint and no new safety surface: composition over the already-guarded `getTransport`, `getRevisions`, `getRevisionSource`.

## Verification

- 4925 unit tests pass; 26 new in `transport-diff.test.ts` plus parser, cosmetics and handler cases.
- Fixtures are real ADT payloads captured from a4h, not hand-written.
- Integration test passes **live** against transport `A4HK906291`.
- Ratchets: two tool-schema budgets and two file-size budgets raised deliberately, each with an inline justification.

End-to-end on the live system:

```
CLAS ZCL_ARC1_DEMO_CALC
  main   exact-transport  prior-revision  00001 (A4HK906289) -> 00002 (A4HK906291)  +1/-0
```

Evidence, SAP's decompiled algorithm, and the plan-vs-reality corrections: [`docs/plans/2026-08-03-transport-diff.md`](docs/plans/2026-08-03-transport-diff.md).


marianfoo:
## Live verification matrix

Built two purpose-made transports on a4h covering create → release → modify-subset → release, with varied object types. **It found two more real bugs, both pre-existing, neither visible from fixtures.**

**A4HK906369** creates `INTF`, `CLAS` (with a CCIMP local class), `PROG`, `DDLS`, `DOMA`.
**A4HK906371** modifies only the class main, the class CCIMP, and the program.

| Case | Observed |
|---|---|
| Created objects, transport still open | ✅ `exact-transport` / `no-prior-snapshot` — and 4 of 5 matched on the **task** id, so the request-only match would have failed |
| Created objects, after release | ✅ same |
| Modified class main | ✅ `prior-revision` 00001 → 00002, exactly the changed line |
| **Modified class CCIMP** | ✅ its own `prior-revision` pair — the multi-include claim, proven |
| Untouched class includes | ✅ listed, `not changed by this transport` |
| Untouched objects (INTF/DDLS/DOMA) | ✅ absent from transport B |
| Metadata type (DOMA) | ✅ inventory, not an error |

Diff B, verbatim:

```diff
### CLAS ZCL_ARC1_TD_CALC [main]
-    rv = iv_a + iv_b.
+    rv = iv_a + iv_b + 1.  " TD-B: off-by-one on purpose

### CLAS ZCL_ARC1_TD_CALC [implementations]
-    rv = iv * 2.
+    rv = iv * 3.  " TD-B: local class changed

### PROG ZARC1_TD_REPORT [main]
-DATA lv_x TYPE i VALUE 1.
+DATA lv_x TYPE i VALUE 42.  " TD-B
```

### Bug 1 — DDLS/DCLS versions endpoint (commit 2)

The CDS view came back `baseline-unavailable`. Probing every type in `revisionsUrlFor`:

| Type | `…/source/main/versions` | `…/versions` |
|---|---|---|
| PROG, INTF, INCL, FUNC, BDEF, SRVD | ✅ | 404 |
| **DDLS, DCLS** | **404** | **✅** |

Not a guessable pattern — `ddic/srvd/sources` uses the long form while `ddic/ddl/sources` and `acm/dcl/sources` use the short one. ARC-1 used the long form for all three, so **every CDS view and access control in a transport reported `baseline-unavailable`**, and revision-id diffs for those types never worked at all. Pre-existing; this PR is just what surfaced it.

### Bug 2 — `99999` is the inactive draft (commit 2)

A CDS view created but not yet activated has *only* a `99999` entry. SAP's `Revision` maps `00000` → "Active" and `99999` → "Inactive". Either can legitimately be the **current** side of a review, but neither is ever a baseline — the predecessor walk now skips both, and the fallback prefers a real snapshot over either.

### CI

13/13 green, including the live `integration` and `e2e` jobs. The two lockfile commits at the base come from #672 (audit advisories, unrelated to this work) and drop out once that merges.

PR #676, opened 2026-08-04
feat: IDE integration — clickable IDE links, roots awareness, stateful-session release

marianfoo:
## What

Two features that bring ARC-1 and the IDE closer together, plus a session-hygiene fix found along the way.

**`ARC1_IDE_LINKS`** — appends a clickable IDE deep link to results that name exactly one object.
`off | auto | vscode | eclipse | <template>`, default `auto`.

**IDE roots awareness** — ARC-1 asks the client once per session what workspace folders it has open, and
uses an `abap:` root to decide whether a link can actually be opened.

**Stateful-session release** (`src/adt/http.ts`) — ARC-1 opened stateful ADT sessions and never released
them, so its session held an ABAP mode until timeout. Eclipse ADT and `abap-adt-api` both send the
release; now ARC-1 does too.

## Why it works this way

Everything here was measured against VS Code 1.131, Claude Desktop, Claude Code and a4h rather than
inferred. The interesting findings:

- **VS Code hands `abap:` workspace folders to MCP servers as roots** — it filters roots by authority,
  not scheme. So an `abap:` root proves SAP's ABAP extension is live *and* a package is open, which is a
  far better gate than the client's self-reported name. Without it, a VS Code user with no ABAP folder
  would get a link that cannot resolve.
- **The root also supplies `{sid}`** (`abap:/repotree-v1/A4H2023` → `A4H2023`) — the only place a system
  id exists on a plain on-prem connection, since `destinationName`/`targetId` are BTP/multi-target only.
- **Claude Desktop renders `vscode://` links but will not open them** (it only hands http/https to the
  OS), so `auto` emitting nothing there is correct, not a gap. Two workarounds were prototyped and
  rejected — see the plan doc.
- **The link is its own MCP content block.** Concatenating it onto the payload made every JSON tool
  result unparseable; a dispatch test now pins the payload block as byte-identical.

## Safety

- `clientAgent` is client-controlled, so it selects a *link format* and nothing else.
- Roots are strictly additive: capability-gated, 2s timeout, per-session cache that also caches failure,
  and impossible on the HTTP transport — every path works without them.
- The link never fails a tool call: errors are swallowed, and `objectBasePath` throwing (FUNC has no
  single addressable path) costs only the `{uri}` placeholder.
- No schema change — the tool-definition fixtures are untouched.

## Docs

`docs/plans/ide-integration-v1.md` plus two research dossiers record what was measured, what was
rejected and why, and the open question about the IDE logouts (hypothesis disproven, candidates listed).

## Test

4996 tests / 172 files, typecheck, lint, action-policy and schema budget all green. Verified live against
a4h: link emission per client, and create/update/delete cycles exercising the session release.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -75,13 +75,13 @@
     "sap_allow_data_preview": {
       "type": "boolean",
       "title": "Allow Table Data Preview",
-      "description": "Allow SAPRead(type=TABLE_CONTENTS) and SAPQuery table preview. Can expose application data — review your data-governance rules.",
+      "description": "Allow SAPRead(type=TABLE_CONTENTS) and SAPQuery table preview. Can expose application data \u2014 review your data-governance rules.",
       "default": false
     },
     "sap_allow_free_sql": {
       "type": "boolean",
       "title": "Allow Freestyle SQL",
-      "description": "Allow read-only ad-hoc ABAP SQL via SAPQuery. Can expose application data — review your data-governance rules.",
+      "description": "Allow read-only ad-hoc ABAP SQL via SAPQuery. Can expose application data \u2014 review your data-governance rules.",
       "default": false
     },
     "sap_allow_transport_writes": {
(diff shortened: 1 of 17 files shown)


marianfoo:
closed for now, reopen later in october

PR #682, opened 2026-08-05
docs: add clean core integration guidance (SAP Note 3690029)

marianfoo:
## Why

SAP's [*Clean core integration* white paper](https://www.sap.com/documents/1327955) and the underlying [SAP Note 3690029](https://me.sap.com/notes/3690029) define an **A–D level scale for integration technologies** — separate from the extensibility scale (Note 3578329) that `sap-clean-core-atc` already uses. Same letters, different axis. Nothing in the repo referenced 3690029 before this.

Most of the white paper is platform-side (Integration Suite, LeanIX, Cloud ALM, ICoE governance) and out of ARC-1's reach. This PR takes only the parts an ABAP system can actually answer.

## What

- **`sap-migration-dossier`** — new *Integration Interfaces* section: the note's level table restricted to ABAP-visible technologies, the evidence calls with explicit column allowlists, scoping and gating rules, and a canonical interface record so the level distribution counts interfaces rather than artifacts. Runs only when the user asks about interfaces; the default custom-code path is unchanged. The review schema splits into `extensibilityLevel` and `integrationLevel` so the two A–D scales can never be merged.
- **`sap-clean-core-atc`** — names its axis explicitly, and maps the `cleanCoreLevel` it receives from mcp-sap-docs onto `extensibilityLevel`.
- **`sap-api-policy-and-architecture`** — §1's "two questions, often conflated" argument now cites a primary source: the note itself says clean core and the API policy are *"two independent, orthogonal governance dimensions"*. §3.2 records that §2.2.2 is restated in the white paper (p. 26) and applied at technology level in the note (JDBC/ODBC mass extraction, ODP-RFC forbidden). **ARC-1's position is unchanged** — the clause is simply propagating into SAP's operational guidance.

## Also included: a dependency fix

`npm audit --audit-level=high` was failing on **every** branch including `main`, not just this one: `nanoid < 3.3.17` (GHSA-2v37-7h3g-55p8, high) is reachable as a dev-only transitive via `vitest → vite → postcss`. Since `test` is a required check, nothing could merge until it was cleared, so the lockfile bump to 3.3.18 rides along here.

Lockfile-only, matching the approach in #672, and **dev-only** — the published npm package and Docker image are unaffected, which is why this PR stays typed `docs:` and cuts no release. Applied with npm 11 rather than the local npm 10, which rewrites lockfile metadata; the resulting diff is the single `nanoid` entry.

## Verification

**Live against a4h (S/4HANA 2023), via `SAPQuery`:**

| Source | Result |
|---|---|
| TADIR census | 46 object types in one 52 ms query — the whole interface surface |
| `TFDIR` (`FMODE = 'R'`) | 9 custom RFC-enabled FMs; `/AIF/PL_BAL` → `/AIF/SAPLPL_BAL` confirms namespace-first main programs |
| `EDBAS` / `EDP13` | 1 custom IDoc type; 1 outbound partner profile bound to a live port |
| `EDIPORT` + domain `EDI_PORTYP` | the authoritative port-type map (`1` tRFC, `3` File, `7` XML HTTP, …) — replaces guessing which `EDIPO*` table holds a port |
| `DD02T` on `EDIPO*` | `EDIPOF` is "Ports for ABAP programming interface", **not** a file port; `EDIPOD` is the file one |
| `RFCDES` + `RFCTYPE` domain | `G` = HTTP to external server, `H` = HTTP to ABAP system, `W` = WebSocket RFC |
| `DD03L` on `RFCDES` | confirms `RFCOPTIONS`/`RFCOPTION1..V` alongside the two fields needed — hence the column allowlist |
| `SAPRead(type="API_STATE")` | contracts + state per object |

**Code-verified, not live-verified:** the `TABLE_QUERY` path (`data` scope, `LIKE`/`IN` in `ALLOWED_OPS`, 10k clamp, `SELECT *` when `columns` is omitted, `AND`-joined where-clauses). Both instances available here run `allowFreeSQL=true` with `allowDataPreview=false`, so `TABLE_QUERY` is refused on them — which is itself why the section documents both gates as independent and gives a no-gate path.

Levels are quoted from the note rather than derived, including its own caveat that the level for a given scenario is use-case dependent. Technologies ARC-1 cannot see (SOA Manager, ABAP proxy via XI, SFTP, JDBC/ODBC, BW extractors, SLT, Delta Share) are listed as evidence gaps rather than omitted or guessed.

No source changes, no tool-surface change. `npm test` (4969 passed, 170 files), `npm run typecheck`, `npm run lint`, and both `npm audit` gates clean locally.


marianfoo:
Applied all five review findings — all correct, thanks. Pushed in 3856f6bb.

**1 (P1) — scope/gate error.** Confirmed: `SAPQuery` is `sql` scope and is only registered when `allowFreeSQL` is set (`policy.ts:78`, `tools.ts:1036`), so the documented `data` + `SAP_ALLOW_DATA_PREVIEW` was wrong, and routing everything through `SAPQuery` also broke this skill's own "no `SAPQuery` unless free SQL and deep evidence" rule. All simple reads now use `SAPRead(type="TABLE_QUERY")` — `data` scope, structured `where`, `LIKE`/`IN` both supported (`ALLOWED_OPS`), 10k row cap. `SAPQuery` is demoted to an optional deep-evidence optimization for the grouped census.

Worth recording: **the two gates are independent in both directions.** Both of my live instances have `allowFreeSQL=true` and `allowDataPreview=false` — `TABLE_QUERY` is refused there while `SAPQuery` works. So the section now names both gates and gives a `tadir_lookup` fallback when neither is available. Consequence for this PR: the `TABLE_QUERY` path is code-verified, not live-verified.

**2 (P2) — scope.** Fixed. Filters derive from the resolved dossier scope (prefixes, namespaces, object list, `DEVCLASS`); the `Z%` literal survives only as an illustrative example, marked as substitutable. `EDIPO*`/`EDP13`/`RFCDES` are now flagged as system-wide configuration that needs explicit approval and is reported as landscape context, not scoped inventory.

**3 (P2) — IDoc B vs C.** Correct, and the investigation turned up more than the finding stated. The join now goes `EDP13.RCVPOR` → port, exactly as you said. But the port *type* cannot come from `EDIPO.PORTTYP`: on a4h `EDIPO` is **empty** while `EDP13` references a live port `A000000001` that is actually defined in `EDIPOA` (tRFC, `LOGDES` = destination). `EDIPOXH` is the XML/HTTP counterpart. So the rule is now "resolve the type by which `EDIPO*` table defines the port", with `EDIPO.PORTTYP` explicitly called out as unreliable. Without the full join, IDocs are reported **B/C indeterminate** and counted into neither level.

**4 (P2) — KPI mislabel.** Correct. The white paper names its KPI 1 "upgrade stability", but the note is explicit that upgrade stability is excluded from the level evaluation because it depends on the underlying object's release state — the note wins. Now called an integration-technology level distribution, with upgrade stability reported separately from `API_STATE` / Business Accelerator Hub.

**5 (P2) — Level D evidence.** Split per technology. SFTP and JDBC/ODBC are marked not ABAP-detectable (ask, don't infer), and the 7.51+ requirement on `source_code` search is called out so a clean result on older ECC reads as "not searched" rather than "not present".

`npm test` 4969 passed, lint clean.

marianfoo:
Second round applied in 0692082b — all five correct, plus the PR description fixed.

**1 (P1) — impossible fallback.** Confirmed: `SAPSearch.tadir_lookup_db` / `_both` are `sql` scope + `FreeSQL` opType (`policy.ts:74-75`), so the fallback recommended the one thing that cannot work when free SQL is off — and `handleSAPSearch` requires `names[]`, so it can't discover by type anyway. Replaced with best-effort `SAPRead(type="DEVC")` per package plus an ordinary object search, and an explicit instruction to mark the TADIR census, legacy SEGW types and RFC/IDoc/destination coverage as **unavailable**.

**2 (P1) — `SELECT *` on config tables.** Confirmed at `client.ts:274`. Every evidence row now carries an explicit `columns` allowlist, and `RFCDES` is restricted to `RFCDEST, RFCTYPE`. Checked the actual field list on a4h via `DD03L`: `RFCDES` also holds `RFCOPTIONS` and `RFCOPTION1..9/A/B/T/U/V` — the packed connection string including logon data. The skill now names those fields and forbids selecting them. Good catch; this was the one finding with a security consequence rather than an accuracy one.

**3 (P2) — cap and AND semantics.** Confirmed `where` clauses are `AND`-joined (`client.ts:295`), so two `LIKE`s on one field match nothing. Now: one query per prefix/namespace, union client-side, and a partition returning **exactly** 10,000 rows is reported as truncated rather than counted.

**4 (P2) — TFDIR package scope.** Right, an FM name doesn't encode its package. Name-prefix scopes may filter `FUNCNAME`; package dossiers resolve the package's `FUGR` objects and match `TFDIR.PNAME` (`SAPL<group>`) before checking `FMODE`.

**5 (P2) — XML/HTTP ≠ HTTPS, and XI proxy.** Both right, and `DD03L` on `EDIPOXH` backs the first: its columns are `PORT, LOGDES, VERSIO, CONTENTTYPE, VERSION, SOAP_PROTOCOL` — the port only *names* a destination, and TLS is configured there. Since that SSL setting lives inside the `RFCDES` blob finding 2 just told us not to read, the resolution is to stop rather than dig: XML-HTTP ports are **candidate** Level B pending SM59 confirmation, with an explicit "do not parse `RFCOPTIONS` to find out". ABAP proxy via XI is now its own row marked as no-ADT-evidence.

**Metadata.** You're right that the description contradicted the follow-up comment. Rewritten: it now separates live-verified `SAPQuery` results from the **code-verified only** `TABLE_QUERY` path (both instances here run `allowFreeSQL=true` / `allowDataPreview=false`), and corrects "no ports configured" — `EDIPO` is empty, but `EDP13` binds to a live tRFC port in `EDIPOA`, which is exactly what made `EDIPO.PORTTYP` unusable.

`npm test` 4969 passed (170 files), lint clean, `git diff --check` clean. Dependencies: `npm ci` in the worktree if you want to rerun the focused files.

marianfoo:
Third round applied in d146eac6. All five verified against the live data dictionary rather than assumed.

**1 — namespaced function groups.** Correct. Verified on a4h: `/AIF/PL_BAL` → `/AIF/SAPLPL_BAL`. The namespace precedes `SAPL`, so the old rule returned zero rows for every namespaced group while looking like it worked.

**2 — `EDIPOF`.** Correct, and it prompted a better fix than a relabel. `DD02T` confirms `EDIPOF` = "Ports for ABAP programming interface" and `EDIPOD` = "Table for File Port Definitions". But the same query surfaced **`EDIPORT` — "Summary Table for all Port Types for IDoc Processing"** — so the whole "which `EDIPO*` table holds the row" heuristic is gone, replaced by one `EDIPORT.PORTTYP` lookup with the codes read from domain `EDI_PORTYP`: `0` CPI-C, `1` tRFC, `2` SAPcomm, `3` File, `4` Internet, `5` ABAP programming interface, `6` XML File, `7` XML HTTP. The live port on a4h reads `PORTTYP = 1`, consistent with it living in `EDIPOA`.

**3 — destination type `G`.** Correct. Pulled the `RFCTYPE` domain: `G` = "HTTP Connection to External Server", `H` = "HTTP Connection to ABAP System". `G` was missing and is the outbound-to-non-SAP case that matters most here. Also added `W` (RFC via WebSockets), which the level table already referenced.

**4 — counting unit.** Fixed. The section now defines a canonical **interface record** (`{id, technology, level, evidence[]}`); one RAP service is `SRVD`+`SRVB`, one IDoc scenario is basic type + partner profile + port. The level distribution counts records; raw artifact counts are reported separately and never summed with them.

**5 — review schema.** Split into `extensibilityLevel` (3578329) and `integrationLevel` (3690029, incl. `FORBIDDEN` and `indeterminate`), in both `SKILL.md` and `references/auditor-patterns.md`. One nuance: `sap-clean-core-atc`'s `cleanCoreLevel` is a field returned by mcp-sap-docs' `sap_get_object_details`, not ours to rename — so it now carries a note that it is the extensibility axis and maps to `extensibilityLevel` on a dossier card.

---

**On CI** — the failure is real but not from this diff, and the split is slightly different from your read of the earlier run. On the current run: `test (24)` fails at the **Security audit (npm audit)** step; `test (22)` reports "fail" but its conclusion is `cancelled` — matrix fail-fast, no timeout involved.

The advisory is `nanoid@3.3.16` (< 3.3.17, high), reached only as a dev transitive: `vitest → vite → postcss → nanoid`. It is present on `main` (currently 9c29f6d3), so it blocks main and every open PR equally. Since `test` is a required check, **this PR cannot merge until that is fixed**, and the fix belongs in its own `fix(deps)` PR rather than bundled into a docs change (precedent: #672).

One caution for whoever takes it: `npm audit fix` on npm 10 silently strips `libc` from `package-lock.json` (this machine is on 10.9.4). Do the bump on npm 11, or diff the lockfile key-by-key against main before pushing.

`npm test` 4969 passed (170 files), lint clean, `git diff --check` clean.

PR #706, opened 2026-08-18
fix: forward user=* when listing transports across owners

FinkeFlo:
## Problem

`SAPTransport(action="list", user="*")` documented an all-owner query, but ARC-1 removed the `user` parameter before calling ADT. SAP therefore applied its default owner filter and returned the authenticated SAP user's view.

```ts
// Old behavior: user="*" was omitted
if (user && user !== '*') {
  params.set('user', user);
}
```

## Root cause

The special case is on the client side, not in SAP:

- On both SAP_BASIS 7.50 and 7.58, the CTS ADT controller reads the optional `user` query parameter and defaults an omitted/initial value to `sy-uname`.
- The downstream CTS request selection treats an owner containing `*` or `+` as ABAP range option `CP`; otherwise it uses `EQ`.
- Independent ADT clients also pass the selected user value through instead of removing `*`.

So omission and `user=*` are deliberately different requests.

## Fix

- Forward every non-empty `user` value, including `*`.
- Keep SAP's returned order; transport numbers are not last-change timestamps, so the unrelated "newest-first" ID sort was removed.
- Clarify that `*` means all owners visible to the authenticated SAP identity.
- Add low-level and handler-level regression tests that parse the outgoing URL and assert `searchParams.get("user") === "*"`.
- Improve the truncation hint for wildcard queries.

```ts
if (user) {
  params.set('user', user);
}
```

## Live verification

The built client from this PR was tested with the exact `user=*` value:

| System | SAP_BASIS | Default/omitted user | `user=*` | Result |
|---|---:|---:|---:|---|
| A4H | 7.58 | 66 | 1015 | wildcard is a superset |
| NPL | 7.50 | 6 | 7 | wildcard is a superset |

The currently deployed pre-fix server also reproduced the bug: exact `user="*"` returned the same counts as omission, while the equivalent pattern `user="**"` returned the larger sets.

## Validation

- Focused transport regression suite: 238 tests passed
- Full unit suite: 5,309 tests passed
- Typecheck, lint, action-policy validation, file/schema budgets: passed
- Root and AppRouter npm audits: no vulnerabilities
- Build, npx package smoke, and MTA validation: passed
- `git diff --check`: passed

## Payload note

`status="D"` is sent to SAP and also enforced client-side. `summary=true` and `maxResults` reduce the returned MCP payload, but only after ARC-1 has fetched and parsed SAP's complete matching CTS response; they are not server-side row limits. On systems with large transport backlogs, callers should prefer a specific owner when possible.

### DIFF
--- a/src/adt/transport.ts
+++ b/src/adt/transport.ts
@@ -136,12 +136,14 @@ export async function listTransports(
 ): Promise<TransportRequest[]> {
   checkTransport(safety, '', 'ListTransports', false);
 
-  // Build query params following sapcli's pattern:
-  //   user={user}&target=true&requestType=KWT&requestStatus=DR
+  // Build query params.
   // requestType=KWT covers Workbench, Customizing, Transport of Copies.
   // requestStatus is sent server-side; we also filter client-side as a fallback.
+  // user=* is sent as-is; URLSearchParams encodes it to user=%2A which SAP interprets
+  // as "all users" (system-wide organizer view). Omitting the param returns only the
+  // authenticated user's requests — which is NOT the same as all users.
   const params = new URLSearchParams();
-  if (user && user !== '*') {
+  if (user) {
     params.set('user', user);
   }
   params.set('target', 'true');
--- a/src/adt/transport.ts
+++ b/src/adt/transport.ts
@@ -139,9 +139,8 @@ export async function listTransports(
   // Build query params.
   // requestType=KWT covers Workbench, Customizing, Transport of Copies.
   // requestStatus is sent server-side; we also filter client-side as a fallback.
-  // user=* is sent as-is; URLSearchParams encodes it to user=%2A which SAP interprets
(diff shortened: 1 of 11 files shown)


FinkeFlo:
Hi @marianfoo 👋

This PR fixes a bug where `SAPTransport(action="list", user="*")` returned only the authenticated user's own transport requests instead of all users' requests system-wide.

**Root cause:** When `user="*"` was passed, the `user` parameter was omitted from the ADT CTS request. SAP then scoped the response to the calling user. Sending `user=*` explicitly returns all users' transports.

**Live-verified** against a SAP development system — with `user=*` sent: 400+ requests system-wide; without: only own requests.

**Also includes:** sort newest-first before the `maxResults` cap, and an improved truncation hint for the `user="*"` case.

All 5310 unit tests pass locally. Happy to address any review feedback!

PR #709, opened 2026-08-20
docs: route deploying agents to the BTP runbook and guard shared services

marianfoo:
## Why

Most users deploy ARC-1 to BTP CF by cloning this repo and asking an LLM for help. That LLM loads `CLAUDE.md` → `AGENTS.md`, which is entirely contributor-facing: it never routed a deploy task to [`docs_page/btp-cloud-foundry-deployment.md`](docs_page/btp-cloud-foundry-deployment.md), and its most prominent deployer-relevant content is the `SAP_*` env table — the wrong surface on CF.

The docs were already good. They were unreachable from where the agent starts.

## What

**1. `AGENTS.md` — deploy routing**

New `## Deploying ARC-1 (not developing it)` section before `## Configuration`: routes to `btp-overview` → the CF runbook, plus five traps (`.env` vs `.mtaext`, override-never-removes, shared services, greenfield-vs-redeploy, human handoffs). The "copy `.env.example`" line is qualified as local/stdio/Docker.

**2. Shared-service guardrail**

`arc1-destination`/`arc1-connectivity` are `org.cloudfoundry.managed-service` and are named after their *resource*, so a space with differently-named instances gets additional instances rather than reuse. Because ARC-1 uses **subaccount-level** destinations, the duplicate normally costs `lite` quota (the deploy fails at the limit) and landscape clarity rather than function — it only loses targets when the shared instance carries instance-level destinations. An extension cannot change a resource `type`, only repoint it with `service-name:`, which leaves MTA owning the instance.

**3. `.env*` now actually means `.env*`**

`mta.yaml` `ignore:` and `.cfignore` matched `.env` and `.env.*`, so `.envrc` was being packaged into the deployed payload. Both widened to `.env*`, with the descriptor test updated in lockstep. The AGENTS wording now correctly names the MTA module `ignore:` as the archive filter — `.cfignore` governs direct `cf push` staging and is *not* a second MTAR filter.

**4. The MTAR secret gate now inspects the payload and fails closed**

The runbook called archive inspection a release gate, but the command listed only the wrapper. Each module is a nested `<module>/data.zip`, and the UI variant ships more than one. The gate now enumerates every member, inspects each, and returns zero only after all of them pass — no archive, no `data.zip` member, a failed extraction, or an unreadable payload are all errors.

**5. `.gitattributes`**

`text=auto eol=lf` so Windows checkouts stop tripping Biome (its formatter `lineEnding` defaults to `lf`). `tests/fixtures/abap/** -text` protects 4 fixtures that mirror real SAP source with CRLF and are read back as source by tests.

## Verification

The gate was extracted verbatim from the rendered doc and run:

| Case | Result |
|---|---|
| Real two-payload UI MTAR, clean | 647 + 3 entries → exit 0 |
| No `mta_archives/` | FAIL → exit 1 |
| MTAR with no `data.zip` member | FAIL → exit 1 |
| Planted `tls.pem` in the *second* of two payloads | FAIL → exit 1 |
| Corrupt payload | FAIL → exit 1 |
| `.envrc` planted in the working tree | absent from both payloads |

Plus `npm run lint`, `npm run btp:validate`, the `mta-descriptor` test, `mkdocs build --strict`, and `git add --renormalize .` proven a no-op with ABAP fixtures round-tripping byte-identical.

Independently reviewed on Windows (PowerShell 5.1, renamed-module fork) and by a third reviewer who clean-room tested the routing premise from `origin/main` before reading the diff.

## Not in scope

A local `npm run btp:preflight` validating a merged `.mtaext` against the real `validateConfig` was designed and deliberately deferred — no deployment on record has paid the slow-loop cost it removes. The hole it would close (shared-Basic multi-target needs `instances: 1`, which lives in module `parameters` where `validateConfig` cannot see it) remains open.

Also skipped: a deployment skill (Claude-only; the AGENTS section covers the same ground), `llms.txt` (helps web-browsing agents, not the clone scenario), a separate `DEPLOYING.md` (a second runbook that drifts), and adding an extension-selectable existing-service path (`vcap.destination[0]` is keyed by service label, so two active destination resources would bind nondeterministically).


marianfoo:
## Follow-up: shared-service task resolved as "accept the limitation"

Investigating whether to add an extension-selectable existing-service path turned up two facts that settle it — and that required correcting this PR's own wording (commit `2d3a120`).

**1. The failure was overstated.** ARC-1 uses **subaccount-level** destinations ([btp-destination-setup.md:29](docs_page/btp-destination-setup.md#L29)), and those resolve through *any* Destination service instance in the subaccount. A duplicate instance therefore still finds the same targets. The real costs are Destination/Connectivity `lite` quota — the deploy **fails** on a subaccount at its limit — and landscape confusion. ARC-1 only genuinely loses targets when the shared instance carries **instance-level** destinations, which the docs already discourage.

**2. A paired managed/existing resource design would add a footgun.** `parseVCAPServices` in `@arc-mcp/xsuaa-auth` reads `vcap.destination[0]` — keyed by service **label**, not instance name. Two simultaneously-active destination resources would bind two instances and pick the first nondeterministically. Preventing a narrow, already-discouraged configuration by introducing a mutually-exclusive resource pair plus a guard against activating both is a bad trade.

So: no descriptor change. The limitation is documented accurately instead, cross-linked to the visibility section. The `existing-service`-in-`mta.yaml` trailer was also dropped from the AGENTS row — it was accurate but read as a recommended route to something we deliberately do not offer.

Worth re-checking in review: the claim that subaccount-level destinations resolve through a newly created instance is from the project's own destination docs, not from a live test on a landscape with two instances.

marianfoo:
## Review response — 2 of 3 fixed, 1 was already fixed on a newer head

Thanks — finding 2 was a genuine defect I could not have caught (no Windows, no built MTAR), and chasing it turned up a worse one in the command it was modelled on.

### Finding 2 (PowerShell MTAR listing) — fixed, and the bash form was worse

Confirmed: `Get-Item mta_archives\arc1-mcp_*.mtar` returns an array once more than one version has been built, and `OpenRead()` takes one path.

Testing the bash original found the same latent bug with a nastier failure mode. With a two-match glob, `unzip -l` treats the second path as a filter *inside* the first archive:

```
$ unzip -l arch/pkg_*.mtar
Archive:  arch/pkg_1.0.0.mtar
        0                     0 files
exit=11
```

A release gate that reports an empty archive is a false green — strictly worse than your loud `OpenRead` error. Both forms now resolve the newest archive explicitly and echo which one they inspected, and the note says why. The PowerShell form also disposes the handle, since an open `ZipArchive` locks the file and the next build fails. Fixed in `604a68b1`; the bash form is tested against a two-archive directory. **The PowerShell form is still untested by me** — it mirrors the shape you ran successfully, but please re-run it.

### Finding 1 (empty Destination catalog) — already fixed; you reviewed a stale head

You fetched `62a2d932`. Head is `2d3a1208`, whose whole purpose was this: I reached the same conclusion from [btp-destination-setup.md](docs_page/btp-destination-setup.md#destination-level-and-visibility) and rescoped the claim before your review landed. It now says the duplicate costs `lite` quota (deploy **fails** at the limit) and landscape clarity, and only loses targets when the shared instance carries instance-level destinations.

Your live evidence is a better corroboration than I had — existing `arc1-destination`/`arc1-connectivity` bound to both apps, and the LLM nearly adding `arc1-multi-destination`/`arc1-multi-connectivity`. That confirms the guardrail is worth having even with the consequence corrected.

Separately, the fact that your fork references those two as **existing services** in its own `mta.yaml` is exactly the case the deferred descriptor task covers — decided as "accept the limitation" in the comment above, and your setup is the concrete counter-example to weigh if it is revisited.

### Finding 3 (`.env` too categorical) — adopted

Reworded to your framing in `604a68b1`: what is unconditional is that `.env` values never reach CF; "healthy app with no SAP target" is the outcome against the target-free base.

### CI

You could not read it (no `gh`, GHE-scoped MCP). All green on `2d3a1208`: `test (22)`, `test (24)`, `mta-validate`, `dependency-review`, CodeQL ×3. `MERGEABLE`.

marianfoo:
## Thanks — the approval evidence exposed a bigger defect than the one it verified

Your entry listing is the useful part:

```
arc1-multi-mcp-server/data.zip
```

The whole application is one **nested zip**. I reproduced it upstream by building 1.1.0: 1.4 MB of the 1.45 MB archive, **648 files**, and the six-entry listing shows none of them.

So the release gate could not see what the paragraph directly under it claims it verifies — *"the application payload must not contain `.env*`, `.npmrc`, service-key exports, private keys, certificates…"*. Both shells listed the wrapper and reported clean. That is pre-existing, not something this PR introduced, but it made my two fix commits polish a command that could not do its job.

Fixed in `8df6a6c4`. The bash form now extracts the nested payload and greps it, and says to read the full listing too, since the gate exists to catch what no denylist anticipated. Run verbatim on the real archive:

```
mta_archives/arc1-mcp_1.1.0.mtar
6 files                          <- wrapper
648 files                        <- payload
PASS: no secret-shaped entries
```

Also excluded `.gitattributes` from the module build — this PR added it and it was landing in the deployable payload next to the other ignored dev dotfiles.

### One more Windows ask, sorry

The PowerShell form is rewritten to expand the nested archive, and it is **untested by me** again. `Expand-Archive` requires a `.zip` extension, so it copies first:

```powershell
$mtar = Get-ChildItem mta_archives\*.mtar | Sort-Object LastWriteTime -Descending | Select-Object -First 1
$mtar.FullName
Copy-Item $mtar.FullName "$env:TEMP\mtar.zip" -Force
Expand-Archive "$env:TEMP\mtar.zip" "$env:TEMP\mtar" -Force
Expand-Archive (Get-ChildItem "$env:TEMP\mtar" -Recurse -Filter data.zip).FullName "$env:TEMP\payload" -Force
Get-ChildItem "$env:TEMP\payload" -Recurse -File |
  Where-Object Name -match '\.env|\.npmrc|service-key|\.(key|pem|p12|pfx|pse|jks|keystore)$'
Remove-Item "$env:TEMP\mtar.zip","$env:TEMP\mtar","$env:TEMP\payload" -Recurse -Force
```

Expect ~648 files under `$env:TEMP\payload` and no matches from the last command. If `Expand-Archive` balks on your PowerShell version, tell me and I will fall back to the `OpenRead` + `MemoryStream` form.

### On the rest of your review

Glob, wording, and `.env` verdicts all noted and unchanged. Your fork answer settles the deferred descriptor task: a committed `existing-service` patch that survived an upstream merge with only a header/version conflict, and not painful enough to justify reopening. That stays closed, with your setup recorded as the counter-example if it ever comes up again.

marianfoo:
## Change made — `c50f7fb6`

You're right, and the number was doubly bad: 648 is ZIP **entries** (629 files + 19 directory records), *and* it was already stale — the previous commit in this same PR added `.gitattributes` to the ignore list after I took that measurement.

So I dropped the count rather than correcting it. It tracks `dist/` growth every release and differs per fork; "hundreds of files behind that single entry" is the point the sentence was making. Nothing else in the PR carries a count.

## Confirmed from your run

- **PowerShell block works on Windows PowerShell 5.1** — `Expand-Archive` accepted the copied `.zip`, found the renamed module payload, 0 denylist matches, cleanup OK. No `OpenRead`/`MemoryStream` fallback needed.
- **Your content scan is the strongest evidence this gate has ever had**: 629 files checked for private-key headers and credential assignments, none found; the `cookies.*` hits are compiled ARC-1 source; and 306 source maps carry no `sourcesContent`, so no original TypeScript ships. That last one is worth having on the record — nobody had checked it.

## `.gitlab-ci.yml`

Agreed it's fork-local and shouldn't block this PR. For the record, your own breakdown shows the upstream payload is clean without it — `dist/` (623), `bin/` (2), `package.json`, `package-lock.json`, `LICENSE` — and `.gitattributes` is now excluded, so upstream packages nothing stray. Adding `.gitlab-ci.yml` to your fork's ignore list is a one-line change alongside the ones you already carry.

Head is `c50f7fb6`.

marianfoo:
## All four findings fixed and reproduced — `820dd74b`

This review found a defect in a supported build path that two prior rounds missed, plus a false claim I had asserted from reading config rather than testing it. Every finding reproduced locally.

### High — the gate missed a payload in the UI build ✅

Reproduced. `npm run btp:build-ui-ext` ships two modules:

```
arc1-mcp-server/data.zip
arc1-ui-router/data.zip
```

`unzip -p "$MTAR" '*/data.zip'` concatenated them into an invalid archive whose listing showed only the 3-file AppRouter payload — the 647-entry server payload, the one the gate exists for, went uninspected. PowerShell had the matching array bug via `-Filter data.zip`.

Both shells now enumerate every `*/data.zip` member and inspect each separately. Verified on a real two-payload MTAR:

```
-- arc1-mcp-server/data.zip:      647 entries
-- arc1-ui-router/data.zip:         3 entries
PASS: no denied paths in any payload
```

The wrapper-entry-count assertion is gone rather than re-stated — that's the second count in this PR to rot, so the lesson landed.

### Medium — clean result returned failure status ✅

Right, and it matters precisely because the audience is an agent executing the block. Now exits 0 when clean, non-zero when a denied path is found. Both directions tested — a planted `\.json$` pattern exits 1.

### Medium — the `.env*` claim was false twice over ✅

Both halves confirmed:

- `.cfignore` governs direct `cf push` staging, not the MTAR. Only MBT's module `ignore:` filters the archive, so "stripped by BOTH" was wrong. I asserted that from reading two files that happened to contain similar lists.
- Both lists matched `.env` and `.env.*`, not `.env*`, so `.envrc` was packaged. I planted one and reproduced it.

Widened both lists to `.env*`, updated the descriptor test in lockstep, and corrected the AGENTS wording. Re-verified: `.envrc` no longer appears in either payload.

### Low — AGENTS row had become a second source of truth ✅

Agreed, and your `mta.ResourceExt` point is the sharper version of it — an mbt diagnostic string that can change without the underlying rule changing. Row compressed to the durable invariant plus a link to the runbook preflight.

### On your pass-1/pass-2 test

That was the answer I most wanted and could not get myself. Two things stand out:

- You reached the runbook on `main` *anyway*, via a `rg` sweep and the README — but explicitly noted you avoided the `.env` trap "from independent CF/MTA knowledge and repository search, not from AGENTS.md." That's the honest scope of this PR: it removes a detour and front-loads `mta-overrides.mtaext` and `cf services`, rather than preventing a failure you were going to have.
- Your list of what it still won't do — choose the topology, supply CF/SAP authority, complete Cloud Connector/CERTRULE handoffs — matches what the "Trying to finish it alone" row is trying to set up. Good to have that confirmed from outside.

marianfoo:
## Fail-open gate fixed — `e0dd2cc3`

You're right, and it's the worst of the failure modes found so far: a broken workspace reporting a clean release. `rc` started at zero and an empty loop left it there.

Rewritten as a subshell function that returns zero **only** after inspecting every payload successfully. It now fails on all five paths you listed. Verified by extracting the block **verbatim from the rendered doc** and running it, so what I tested is exactly what ships:

| Case | Result |
|---|---|
| Real two-payload UI MTAR, clean | `647 + 3` entries → **exit 0** |
| No `mta_archives/` | `FAIL: no archive in mta_archives/` → **exit 1** |
| MTAR with no `data.zip` member | `FAIL: no <module>/data.zip member` → **exit 1** |
| Planted `tls.pem` in the **second** of two payloads | `FAIL: denied path in mod-ui/data.zip` → **exit 1** |
| Corrupt payload | `FAIL: cannot read mod-server/data.zip` → **exit 1** |

Two details on your suggestions:

- I used explicit checks inside a **subshell function** rather than a bare `set -euo pipefail`. An operator pastes this into an interactive shell, and a leaked `set -e` would kill their session on the next failing command. The subshell keeps `pipefail` and the cleanup trap contained while still giving one exit status.
- Temp file now lives in an `mktemp -d` directory instead of overwriting `payload.zip` in the working tree, per your note.

PowerShell gets the equivalent: `$ErrorActionPreference = 'Stop'`, explicit member and file count assertions, `try/finally` cleanup. Untested by me as always — the earlier reviewer has been running those.

### Minor cleanups

- "ships two payloads" → "ships more than one payload". That is the **third** count in this PR heading for the same rot; the pattern is now unmistakable.
- The full-listing example no longer hardcodes `arc1-mcp-server/data.zip`. Good catch — the gate went module-agnostic while the example beside it did not, and forks rename the module.
- PR body scope corrected to 7 files.

Nothing else outstanding on my side.

marianfoo:
## This review predates the fix — but it found one more stale thing

The blocking finding and both minor cleanups were fixed in `e0dd2cc3`, pushed before this review arrived. The review text and line numbers match `820dd74b`.

Your exact repro, run against the **remote** head:

```
$ cd $(mktemp -d) && bash <gate extracted from e0dd2cc3>
FAIL: no archive in mta_archives/
documented_gate_status=1
```

Your six requirements, verified against the pushed content:

| Requirement | Status |
|---|---|
| fail if no MTAR is selected | `FAIL: no archive in mta_archives/` |
| fail if no `data.zip` members exist | `FAIL: no <module>/data.zip member` |
| propagate listing/extraction errors | `unreadable archive` / `cannot extract` / `cannot read` |
| inspect every payload | planted `tls.pem` in the **second** of two payloads → exit 1 |
| return zero only after successful inspection | clean two-payload MTAR → 647 + 3 entries, exit 0 |
| unique temp file, not `payload.zip` in cwd | `tmp=$(mktemp -d)` with a cleanup trap |

PowerShell has `$ErrorActionPreference = 'Stop'`, `members.Count -eq 0` and `files.Count -eq 0` assertions, and `try/finally`.

Cited line 237 now reads "ships **more than one** payload" — your wording. Line 264's hardcoded `arc1-mcp-server/data.zip` is gone.

One deliberate deviation: explicit checks inside a **subshell function** rather than a bare `set -euo pipefail`. An operator pastes this into an interactive shell, and a leaked `set -e` would kill their session on the next failing command. The subshell contains `pipefail` and the trap while still yielding one exit status.

## Your "stale summary" point went further than you saw

The PR **body** still carried the `.env*` claim you disproved last round — *"stripped from the MTAR by both `.cfignore` and `mta.yaml` `ignore:`"*. I corrected that in `AGENTS.md` and never in the description, so the PR was still asserting it to anyone reading the top of the page.

Body rewritten to the final state: 7 files, the corrected `.cfignore`-is-not-a-second-MTAR-filter wording, the fail-closed gate, and the full verification matrix. Thanks — that one would have shipped.

Head is `e0dd2cc3`, CI green, `MERGEABLE`. Please re-fetch before the next pass.

PR #719, opened 2026-08-25
fix(read): auto-split nested BSP paths passed only in name

F-Wer:
## Summary
- `SAPRead type=BSP` silently mangled a nested path when the caller put it entirely in `name` (e.g. `name="APP/controller"`) instead of splitting it across `name`+`include`. `getBspAppStructure`/`getBspFileContent` uppercase their `appName` argument in full, so an unsplit `name` got its sub-path uppercased too, producing a 404 against the case-sensitive filestore that reads exactly like "object does not exist" — with no hint to use `include`.
- Auto-split `name` on the first `/` when `include` is not also given, so both call shapes behave the same. No change needed in `client.ts` — `getBspAppStructure`/`getBspFileContent` are correct when called with a clean `appName`.

## Test plan
- [x] Added two regression tests (`tests/unit/handlers/read.test.ts`): subfolder case and file case, both asserting the outgoing request URL preserves the sub-path's original case instead of uppercasing it.
- [x] `npx vitest run tests/unit/handlers/read.test.ts` — 130/130 pass.
- [x] `npx vitest run tests/unit/adt/client.test.ts tests/unit/handlers/schemas.test.ts` — 375/375 pass (both also cover BSP).
- [x] Full unit suite — 4938/4950 pass; the 12 failures are pre-existing and unrelated (POSIX file-permission-mode assertions that don't hold on Windows, plus a few CI-script/workflow tests expecting a CI environment — confirmed by checking an untouched file fails the same lint check for the same repo-wide CRLF reason).
- [x] `tsc --noEmit` — no new errors (pre-existing errors are all missing optional BTP/xsuaa deps, unrelated).

### DIFF
--- a/src/handlers/read.ts
+++ b/src/handlers/read.ts
@@ -744,21 +744,33 @@ export async function handleSAPRead(
             'for the reason (often a missing S_ADT_RES authorization), or set SAP_FEATURE_UI5=on to force it on.',
         );
       }
-      const include = args.include as string | undefined;
+      let include = args.include as string | undefined;
       if (!name) {
         // List all BSP apps (optional search via query param not used here since name is empty)
         const apps = await client.listBspApps();
         return textResult(toolJson(apps));
       }
+      // A caller sometimes puts the whole nested path in `name` instead of splitting it
+      // across name+include (e.g. name="ZAPP/controller" instead of name="ZAPP"
+      // include="controller"). getBspAppStructure/getBspFileContent uppercase their
+      // appName argument in full, so leaving this unsplit would uppercase the sub-path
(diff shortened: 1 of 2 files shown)


marianfoo:
Thanks for reporting and addressing the mixed-case BSP path failure. I independently reproduced the original 404, but the first-slash split in this PR regresses namespace-style BSP applications such as /UI2/USHELL, and the existing dot heuristic still misclassifies dotted folders and extensionless files. #727 supersedes this PR with namespace-aware parsing and response Content-Type classification, validated on SAP_BASIS 7.50 and 7.58.

PR #734, opened 2026-09-01
feat(server): optional system label in the MCP server instructions

### LINKED ISSUE #733: Instances cannot self-identify: the server instructions never say which SAP system an instance reaches
With several ARC-1 instances pointing at different SAP systems (dev/QA/prod), `ARC1_SERVER_NAME` (#606) names each instance in the `initialize` handshake — but some clients replace that name with an opaque connector id in the tool namespace (e.g. claude.ai remote connectors surfaced in Claude Code show `mcp__<uuid>__SAPRead`). What those clients *do* surface verbatim is the server `instructions`, and ARC-1's `SERVER_INSTRUCTIONS` are a static constant: they describe the domain but say nothing about which system the instance reaches ("ARC-1 gives **this SAP ABAP system**...").

The practical effect: a model facing four look-alike ARC-1 servers has to probe each one with a tool call just to find out which is production. We hit exactly this running five instances (dev/QA/prod across two landscapes) side by side.

Proposal: an optional `ARC1_SYSTEM_LABEL` / `--system-label` (free text, e.g. `ERP production (read-only)`) that is prepended as the first line of the server instructions — `Connected SAP system: <label>.` — leaving the handshake name untouched and the default instructions byte-identical when unset. Multi-target mode keeps its own instructions builder.

PR with a possible implementation follows.

MWKAnalytics:
Fixes #733.

Adds an optional `ARC1_SYSTEM_LABEL` / `--system-label`: a free-text label of the connected SAP system (e.g. `ERP production (read-only)`), prepended as the first line of the MCP server instructions — `Connected SAP system: <label>.`

Why the instructions and not just `serverInfo.name`: `ARC1_SERVER_NAME` (#606) covers clients that key on the handshake name, but some clients replace that name with an opaque connector id in the tool namespace (e.g. claude.ai remote connectors surfaced in Claude Code). The instructions are surfaced verbatim there, so the label is what lets a model tell several look-alike ARC-1 instances apart without a probing tool call.

Scope kept deliberately small:

- Default unchanged: with no label the instructions are byte-identical to today's `SERVER_INSTRUCTIONS`.
- Multi-target mode is untouched — it keeps building its own instructions from the discovered targets.
- Config plumbing follows the `serverName` pattern exactly (parser, defaults, CLI flag descriptor).
- Tests: config parsing (default / env / flag-over-env) + `buildServerInstructions` (unchanged without label, label as first line). Docs: configuration-reference, AGENTS.md env table, `.env.example`.

Running in production on five instances (dev/QA/prod across two system landscapes) at our side.


marianfoo:
Superseded by #735, which keeps the model-visible system-label direction from this contribution while adding bounded one-line normalization, CLI support, real MCP initialization coverage, multi-target/default compatibility tests, operator documentation, and the repository's research/plan artifacts.

Thank you @MWKAnalytics (Marvin Kloth) for the original contribution. Marvin is credited as a co-author on the implementation commit in #735. Closing this PR in favor of the replacement.

PR #739, opened 2026-09-02
fix: bound data-preview response memory

### LINKED ISSUE #737: Parallel multi-MB SAPQuery/data-preview responses can OOM the instance - no response-size budget
Running ARC-1 1.1.0 on Cloud Foundry (512M container, `http-streamable`, XSUAA + principal propagation, `SAP_ALLOW_DATA_PREVIEW` + `SAP_ALLOW_FREE_SQL` on): a client issued **two parallel `SAPQuery` calls whose results were each several MB** (free SQL against production BW tables). The instance crashed twice within 31 seconds, both times `Exited with status 137 (out of memory)`:

```
2026-09-01T17:25:55  app.crash  exit_description: APP/PROC/WEB: Exited with status 137 (out of memory)
2026-09-01T17:26:26  app.crash  exit_description: APP/PROC/WEB: Exited with status 137 (out of memory)
```

Each crash killed every in-flight MCP request on that instance; the client (Claude Desktop) surfaced it as "two chunks dropped the MCP connection" and recovered by retrying the same queries sequentially — which fit into memory one at a time.

The operational fix on our side is simply more memory (we bumped the instance to 2G). But two observations that might be worth a server-side guard:

1. `ARC1_MAX_CONCURRENT` (default 10) bounds the **number** of concurrent SAP requests, not the **bytes** they can pin in memory at once. Ten concurrent multi-MB data-preview/SQL responses — each amplified by XML parsing into a much larger object graph before serialization — will blow through any reasonably sized container.
2. A multi-MB tool result is rarely what an LLM client actually wants anyway; it mostly burns context.

Proposal: a configurable per-response size budget for the data-shaped paths (`SAPQuery` table/free-SQL results, data preview), which fails the individual call with a clear, LLM-actionable error ("result exceeds N MB — narrow the selection or page with max_rows") instead of letting the process die and take every concurrent request down with it. A coarse global in-flight-bytes semaphore around the data paths would also do. Happy to contribute a PR if you have a preferred shape.

### LINKED ISSUE #741: mta.yaml pins --max-old-space-size=448 regardless of memory overrides
`mta.yaml` starts the server with a fixed `command: node --max-old-space-size=448 dist/index.js`, sized for the shipped `memory: 512M` with a comment saying "bump in lockstep when memory: changes". Nothing enforces that lockstep, and the documented way for operators to size an instance is an `.mtaext` override of `parameters.memory` — which does not touch `command:`.

We hit exactly that after the OOM in #737: the crashed instance was raised to 1 GiB via `.mtaext`, the heap stayed capped at 448 MiB, and roughly half of the added memory only benefits non-heap allocations. The instance looked "fixed" while V8 still had the old ceiling.

Proposal: derive the flag from the container limit in the start command instead of hardcoding it. Cloud Foundry exposes `MEMORY_LIMIT` (megabytes with an `M` suffix, e.g. `1024M`), so shell arithmetic is enough and there is no dependency on Node's `--max-old-space-size-percentage` (22.21+, above the current engine floor):

```yaml
command: >-
  M=${MEMORY_LIMIT%M}; exec node --max-old-space-size=$(( M * 7 / 8 )) dist/index.js
```

`7/8` keeps today's 448 MiB at 512M byte-identical; whether the ratio should become 75 % (as the #737 research proposes) is a separate decision — this issue is only about the flag following the memory. A one-line startup log of the effective old-space value would make misconfigurations visible in `cf logs`.

Happy to send the PR if you agree with the direction.

marianfoo:
## Goal

Close the data-preview memory-exhaustion path reported in #737 and the related fixed Cloud Foundry heap-sizing issue in #741. This PR keeps the research dossier and implementation together so the chosen limits, tradeoffs, and validation evidence remain reviewable in one place.

## What changed

- Enforces a cumulative 2 MiB post-content-decoding response-body budget per complete data-preview tool call, including internal reads and automatically chunked queries.
- Adds a process-wide FIFO data-result semaphore (default 2), shared across single-target, principal-propagation, pinned, and aggregate clients and held through fetch, XML parsing, result construction, and terminal audit work.
- Caps direct Fetch and explicitly budgeted BTP Connectivity response streams before string conversion; bounds retry bodies, excludes CSRF bootstrap from result accounting, propagates cancellation, and gives the live proxy stream exact client ownership/teardown.
- Returns stable, secret-free `DATA_RESPONSE_TOO_LARGE` errors and emits `data_response_limited` audit events without SQL or response bodies.
- Parses data-preview rows and metrics from one XML parse.
- Clamps raw and chunked `SAPQuery.maxRows` to 10,000 at the sink and reports `rowLimitClamped`, `requestedRows`, and `effectiveMaxRows` when applicable.
- Adds strict positive-integer configuration through `ARC1_MAX_DATAPREVIEW_RESPONSE_BYTES` / `--max-datapreview-response-bytes` and `ARC1_MAX_CONCURRENT_DATA_RESULTS` / `--max-concurrent-data-results`; zero does not disable either boundary.
- Replaces the fixed 448 MiB MTA value with a fail-closed CF launcher that validates `MEMORY_AVAILABLE`, applies the adaptive 75% old-space policy, and `exec`s Node for direct shutdown signals (384 MiB at 512 MiB; 768 MiB at 1 GiB).
- Adds conservative BTP RAM planning from 512M through 4G plus Docker-specific memory/heap guidance, with batch-versus-throughput tradeoffs and durable deployment examples; `AGENTS.md` routes setup work to the BTP table.
- Confirms the earlier red dependency-audit baseline is resolved on current `main` and this branch; both root and BTP AppRouter audits report zero vulnerabilities, and no lockfile delta remains in this PR.
- Extracts bounded-transport, table-query, and runtime-memory helpers so all files pass the existing size ratchets; no file or schema budget was raised.
- Updates tool guidance, operator documentation, release notes, and the implementation dossier.

## Compatibility and operator note

This intentionally changes oversized-result behavior. A query that previously returned more than 10,000 rows is now clamped, and a successful data-preview response that cumulatively exceeds the configured byte allowance fails without partial rows. The effective row count is data-width-dependent and can be much lower than 10,000.

Keep the defaults initially. Approved batch/file consumers may raise the byte ceiling only after measuring bounded peak RSS for their row shapes, normally reducing data-result concurrency proportionally. Use the BTP RAM sizing table in `docs_page/btp-administration.md#data-preview-ram-sizing` to change `parameters.memory` and both data-result limits as one deployment decision, then test the widest approved result at full configured concurrency.

## Validation

- `npm run lint`
- `npm run typecheck`
- `npm run check:sizes` — all 429 tracked source/test files and every tool-schema budget passed
- `npm test` — 186 files, 5,449 tests passed
- `npm run build`
- `npm audit --audit-level=high --omit=optional` — 0 vulnerabilities
- `npm audit --prefix btp/approuter --audit-level=high --omit=optional` — 0 vulnerabilities
- `npm ci && npm test` in `btp/approuter` — 3 tests passed
- `npm run btp:validate`
- `npm run btp:build`
- `npm run docs:build`
- Live, read-only SAP_BASIS 758 test at the shipped 2 MiB ceiling: three parallel `SELECT * FROM TADIR` / 10,000-row calls were all bounded; maximum admitted concurrency was 2; no slot/waiter leaked; a subsequent small query and system read succeeded; incremental peak RSS was 25 MiB.
- Disposable CF Node buildpack 1.9.3 / Node 24.18.0 validation: the launcher produced 384 MiB old-space at 512 MiB and 768 MiB at 1 GiB; restart and scale-down signals reached ARC-1, its shutdown hook ran, and Node exited with status 0.
- Live Destination → Connectivity → Cloud Connector → SAP validation ran three waves of six concurrent oversized Basic-auth queries: every call returned `DATA_RESPONSE_TOO_LARGE`, recovery queries succeeded, and no delayed `ERR_INVALID_STATE` crash occurred. The late-data cancellation regression is also covered locally (old adapter 20/20 failures; fixed adapter 0/20).
- Exact-branch multi-target PP acceptance: OAuth/PKCE login and refresh succeeded; all four per-user targets plus the shared Basic target passed `SAPRead(SYSTEM)`. Six concurrent oversized `A4H/001` PP queries were bounded under the two-slot queue, the recovery query succeeded, and the process stayed healthy. Disposable apps/routes were deleted; existing apps, shared services, and destinations were unchanged.

Detailed rationale and measurements: `docs/research/issues/737-datapreview-response-memory-budget.md`.

Closes #737
Closes #741

Reporter follow-up and comparison: #742.


mta.yaml
```diff
@@ -71,11 +71,11 @@ modules:
...
       health-check-http-endpoint: /health
-      # Node heap target sized to 87.5% of dyno memory (512M -> 448M) so V8 GCs
-      # cleanly inside the CF cgroup instead of being SIGKILL'd; bump in
-      # lockstep when memory: changes.
-      command: node --max-old-space-size=448 dist/index.js
       buildpack: nodejs_buildpack
+      # The wrapper validates the buildpack-provided MEMORY_AVAILABLE, applies
+      # OPTIMIZE_MEMORY's adaptive 75% policy, then execs Node directly so CF's
+      # SIGTERM reaches ARC-1 instead of stopping at npm.
+      command: exec ./bin/start-cf.sh
```
MWKAnalytics: Deployed this branch to our dev instance (BTP, Cloud Connector + principal propagation) with our existing .mtaext, and it crash-looped with exit 126:

```
bash: line 1: /home/vcap/app/bin/start-cf.sh: Permission denied
bash: line 1: exec: /home/vcap/app/bin/start-cf.sh: cannot execute: Permission denied
```

The script is 100755 in git, but we build the MTAR with mbt on Windows, and the archive stores it as 0666 like every other file (checked the zip entry attributes). Invoking it via the shell makes the start independent of the file mode:

```suggestion
      command: exec sh ./bin/start-cf.sh
```

With that one-word change the full checklist from #737 passed on our side (details in the issue). Request changes only because of this case - at least everyone building with mbt on Windows will hit it. Everything else looked good.

MWKAnalytics (changes requested):
Tested the branch (at 95b7fa8c) on our BTP dev instance through Cloud Connector + principal propagation. One blocker (Windows-built MTAR loses the exec bit of the new launcher, suggestion inline), otherwise the full #737 checklist passed - details in #737.

mta.yaml
marianfoo: Addressed in 0a00928b. The MTA now starts with `exec sh ./bin/start-cf.sh`, so the launcher no longer depends on the executable bit preserved by the MTAR builder. Added a regression test that copies the launcher with mode 0666 and verifies the adaptive 512 MiB → 384 MiB startup path. The generated MTAR resolves to the new command; MTA validation/build, 5,450 unit tests, docs, lint, size, and type checks pass locally. Thank you for catching the Windows packaging case.

MWKAnalytics (approved):
Retested at 0a00928b with a fresh Windows-built MTAR, this time without any local patch: the packaged descriptor carries `exec sh ./bin/start-cf.sh`, the instance starts cleanly on our BTP dev space (1/1, both envelope lines in the startup log, no permission error). That was the only blocker from my side - approving.

PR #742, opened 2026-09-02
docs(research): reporter-side MARA/BSEG measurements + heap-flag observation for #737

MWKAnalytics:
Adds the reporter-side evidence discussed in #737 directly to the dossier, so the default-sizing decision can be made against real wide-table numbers rather than the TADIR shape alone:

- **New section "Reporter-side measurements"** (before *Consumer-side budget sizing*): TADIR / MARA / BSEG measured on a customer ERP QA system *through the BTP Connectivity proxy path* the local harness skipped, using ARC-1's own `tool_call_end` `resultSize`. MARA (264 cols) lands at ~0.80 MiB raw XML per 100 rows and hits 1 MiB at ~124 rows; BSEG (331 fields) at ~1.0 MiB per 100 rows (derived — its preview timed out twice at 120 s even with `BUKRS`+`GJAHR`). Conclusion offered: 2 MiB with data concurrency 2 keeps the same 4 MiB envelope with usable headroom on the widest core tables.
- **One paragraph in *Cloud Foundry sizing***: an `.mtaext` `parameters.memory` override does not move the pinned `--max-old-space-size=448` — we ran a 1 GiB instance with a 448 MiB heap after the incident. Tracked separately as #741 since it is a descriptor defect, not part of the response budget.

Docs only, no code. Numbers and method are reproducible from the audit log of any deployed instance.

### DIFF
--- a/docs/research/issues/737-datapreview-response-memory-budget.md
+++ b/docs/research/issues/737-datapreview-response-memory-budget.md
@@ -324,6 +324,33 @@ old-space limit ended in a fatal allocation failure (exit 134). That is not the 
 137 path, but it independently validates that this allocation chain can terminate a process. The
 shared live CF application was deliberately not crashed to reproduce 137.
 
+### Reporter-side measurements (customer ERP QA system, Cloud Connector / principal propagation)
+
+The reporter measured tool-result sizes on a customer ERP QA system through the deployed CF
+instance — i.e. through the BTP Connectivity proxy path the local harness above skips — by reading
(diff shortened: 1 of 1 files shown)


marianfoo:
Closing as superseded by #739, which has now merged. The reporter-side TADIR/MARA/BSEG evidence was incorporated into the #737 dossier with clarified measured-versus-derived units and limitations, and the CF heap-sizing issue from #741 was implemented in the same PR using the validated adaptive launcher. Thank you for providing the production-side measurements and for identifying the heap-flag drift.

PR #743, opened 2026-09-02
fix(read): read methods of class-local classes through their include …

jrodriguez-rc:
…(#1)

SAPRead(type="CLAS", method="lhc_x~method") answered "not found ... Available methods: (none)" for every RAP behavior pool, although the <localclass>~<method> specifier is documented and extractMethod honours it. The method path always fetched source/main, and a behavior pool's MAIN is only the FOR BEHAVIOR OF shell - the handler lives in the implementations include, which was never read. The same condition made method + include together silently ignore method and return the whole include.

The read path now resolves the section the way edit_method already did: an explicit include= wins; else detectLocalHandlerInclude routes lhc_*/ lcl_* to implementations and ltc_* to testclasses; else MAIN. Include reads use client.getClassInclude and bypass the source cache, whose key has no include component. A missing include is reported as such, and a bare name that is not in MAIN gets a hint to qualify it or pass include=. No schema change: include= is already validated by SAPREAD_CLAS_READ_INCLUDES.

Pre-existing since the ts-src -> src rename (2026-04-02); surfaced while reading a RAP behavior pool on SAP_BASIS 816.


marianfoo:
Thanks @jrodriguez-rc — I confirmed the defect and your core source-section routing approach live on SAP_BASIS 816 and 758.

I am closing this PR in favor of #744 so there is one merge candidate. #744 retains your approach and credits you as a co-author of its implementation commit, while also fixing one remaining precedence edge here: explicit `include="main"` is currently collapsed to an omitted include, so a qualified `lhc_*` method is auto-routed to CCIMP instead of honoring MAIN.

The replacement adds the explicit-MAIN regression, the broader local-class/include/cache test matrix, LLM-facing and user documentation, and the live 758/816 evidence. All 11 CI checks on #744 are green, including Node 22/24, dependency review, CodeQL, and SAP integration tests.

Thank you for the precise report and original patch.

PR #748, opened 2026-09-04
feat(write): multi-node Knowledge Transfer Document writes and per-node short texts

jrodriguez-rc:
## Summary

Knowledge Transfer Documents (SKTD/KTD) on a multi-node object — a BDEF with entities, savers, actions, functions and associations — could not be written node by node, and had no way to set the per-node **short texts** ADT shows in its tree. This PR fixes the write and adds the short texts.

### 1. `SAPWrite(type="SKTD", action="update")` writes any subset of nodes (bug fix)

- **Root cause:** `rewriteKtdText` replaced only the first `<sktd:text>` (non-global regex) and had no inverse of the `## <node id>` sections `decodeKtdText` renders, so a multi-node body overwrote the root and silently dropped every other node.
- **Fix:** `rewriteKtdText` is the exact inverse of `decodeKtdText`. Each `## <node>` section is Base64-spliced into that element's `<sktd:text>` (including the self-closing `<sktd:text/>` of a node nobody documented yet); unaddressed nodes stay byte-identical. A body that addresses no node is refused when 2+ nodes carry text (it used to overwrite the root); an empty body is refused rather than erasing documentation.
- **SAPRead** appends a compact index of the nodes SAP pre-created without text (`decodeKtdText` hides them), so a caller can address a new node without first provoking the write's refusal to learn its id.
- The update reads the envelope **without** `version` on purpose: ADT's default view carries the pending inactive draft, so consecutive node writes accumulate instead of reverting to the active version (live-verified).

### 2. Per-node short texts (feature)

- **SAPWrite** `shortTexts=[{node,text}]` on `update` and `create` (SKTD/KTD only): writes `sktd:shortText/@sktd:text` per node, Base64, max 60 UTF-16 units (`sktd:instruction shorttext`), `""` clears. Works with or without `source`. Every assignment is validated — unknown node, duplicate, too long, no `<sktd:shortText>`, `obligation="forbidden"` (root and entity nodes) — **before any lock**; a refusal never leaves a half-applied document.
- **SAPRead** ends with a read-only trailer behind an HTML-comment marker (`<!-- arc1:ktd-meta … -->`): the short text of every node that has one, then the undocumented-node index. `SAPWrite` strips everything from the marker on, so a whole `SAPRead` result pasted back as `source` is a byte-level no-op (pinned by test). A `## ` section pasted *below* the marker is refused, not dropped.
- **One node resolver** for `shortTexts[].node` and `## ` headings: exact id → case-insensitive id → unique node name. `shortTexts[].node` accepts the qualified or bare name, percent-decoded or as encoded on the wire (`ZI_TravelTP.GetPhoto`, `GetPhoto`, `%_OWN`, `%25_OWN`); `## ` headings accept only the qualified spellings, because every BDEF carries `<Entity>.create/update/delete` and `## Update` in the root's prose must stay prose. A heading that is unmistakably a node reference but matches nothing (ADT URI, `#type=` fragment, `<KnownEntity>.<typo>`) is refused rather than folded into the previous node.
- Every name the trailer prints is guaranteed to resolve back to the node it was printed for: a BDEF's root-entity node is named like the object itself and that name resolves to the root, so the entity is listed by its full id.

### Wire facts (all live-verified on SAP_BASIS 816 on-prem; `[E]` evidence, `[I]` inference)

- `[E]` A KTD holds one `<sktd:element>` per documentable node; SAP pre-creates all of them (BDEF: the full behaviour tree; DDLS: root only). Child order `id → text → objectReference → parent → shortText → link`; `<sktd:text>` is Base64 line-wrapped at 76 columns on read.
- `[E]` **SAP derives `adtcore:objectReference/@adtcore:description` from the short text** and clears it with it — ARC-1 writes only `sktd:shortText/@sktd:text`, never the description.
- `[E]` 60 UTF-16 units (30 emoji) accepted by SAP. `[I]` whether SAP itself rejects 62 is untested because ARC-1 refuses first (possibly over-strict for non-BMP text). `[I]` `obligation="mandatory"` never observed live; treated as writable.
- Never synthesize an `<sktd:element>` — the child schema is unconfirmed; ARC-1 refuses instead.

Full wire format, root-cause analysis and evidence: `docs/research/2026-09-02-sktd-multi-node-write.md`.

## Review notes

- **Merge as one.** The SAPRead trailer advertises `SAPWrite shortTexts=[…]`; the read half without the write half would have the schema reject the parameter the trailer recommends.
- **Tool surface:** `shortTexts` added to `SAPWrite` (three-file sync: `tools.ts`, `schemas.ts`, handler; also validated at the schema to SKTD/KTD + update/create). Fixtures regenerated (`vitest -u`, reviewed diff). Budgets moved and annotated in `scripts/ci`: `tools.ts` line ratchet +14; `standard-full-git` schema-token ratchet 17,700 → 17,750; `write-ddic.test.ts` 3000 → 3100 (the SKTD block is the named split). Wire walls untouched.
- **Release review** (7 independent lenses, each finding refuter-verified against HEAD) drove the last round: round-trippable trailer names, qualified-only headings, refusal of a section below the trailer and of a typo in a qualified heading, a heading regex that was quadratic in the line length (ReDoS on LLM-supplied `source`, 1.2 s at 80k chars) made linear; one back-to-front splice helper instead of three copies; the unknown-node refusal lists nodes compactly instead of ~80 raw ids; the SKTD create handler's partial-success guard covers the follow-up GET and PUT.
- **Deferred (follow-ups, not in this PR):** extract `src/adt/ktd-xml.ts` as a move-only refactor (the KTD section is ~45% of `ddic-xml.ts`, now 1206 lines of a 1500 budget); split `tests/unit/handlers/write-ddic.test.ts`.
- `fix:` + `feat:` → next minor. The annotated entry in `docs_page/release-notes.md` goes in while the release-please PR is open, per AGENTS.md.

## Verification

- Unit: `npm test` 5762 passing (9 remaining failures are the known Windows-only environment cases: sqlite/file permissions, deny-actions file paths, skip-discipline, CLI runtime). typecheck, Biome, size ratchets + schema budget, `validate:policy`, build: green.
- Live (on-prem 816): multi-node write of 3 nodes in one call and a 4th separately, refusal before any lock, draft accumulation across consecutive writes; short texts by short name, 60-unit emoji accepted, 62 refused pre-lock, clear with `""`, whole-read paste-back no-op, unaddressed bodies byte-identical; a 91-node BDEF KTD documented in one `SAPWrite` (75 sections + 68 short texts), read back node by node and activated.


marianfoo:
Thank you, @jrodriguez-rc, for the contribution and the thorough live SAP research behind it.

To reduce the review surface and merge risk, we split this work into two focused PRs:

- #749 — safe multi-node KTD long-text routing and round-trip hardening
- #750 — per-node KTD short-text support

Both have now been merged into `main` after additional review, regression coverage, CodeQL, and live SAP integration/E2E verification. Your contribution is retained through co-author attribution on both squash commits.

Closing #748 as superseded by #749 and #750.

PR #772, opened 2026-09-09
feat: batch ATC checks with explicit object coverage

### LINKED ISSUE #770: [Feature]: run ATC on a list of objects in one call
### Preflight

- [x] I searched existing issues (open and closed) and this isn't already requested.

### Problem / use case

Hi Team, thanks for ARC-1! It's been super useful.

SAPDiagnose atc takes one object per run. On our tenant a run costs about 11 s, of which roughly half is the fixed overhead (variant lookup, worklist, run, polling, result fetch), so checking 20 freshly activated objects is the slowest part of our pipeline even with calls in parallel.

A package run (type: "DEVC") works, 101 objects in about 20 s with findings per object, but the package keeps growing and we only want the objects of one batch.

### Proposed solution

Since adtcore:objectReferences is a list on the ADT side, would you consider objects: [{ type, name }] on the atc action, one objectReference per entry?

### Area

SAPDiagnose

### Other area — please specify

ATC

### Release-specific?

_No response_

### Alternatives / workarounds

_No response_

marianfoo:
SAPDiagnose ATC currently repeats worklist setup and polling for every explicit object. This adds `objects: [{type,name}]` for one native selection of up to 20 entries while preserving existing single-object and package calls.

SAP can complete a worklist without reporting every requested object. Batch results therefore include ordered object coverage, nullable finding counts, finding ownership and separate run provenance. A complete initial run may trigger one verification batch containing only unreported objects, using the same confirmed variant, SAP identity and original deadline. Missing, malformed or contradictory evidence stays incomplete. Recoverable verification failures preserve initial findings; cancellation, safety failures and programming errors propagate.

Supported types: CLAS, INTF, PROG (on-premises only), FUGR, DDLS, DCLS, BDEF, DDLX, SRVD, SRVB, **TABL, DTEL and DOMA**. DDIC uses ATC R3TR references, including TABL/DT tables and TABL/DS structures, without metadata/subtype lookups. Known editor and exact source-resource URI equivalents are recognized. BTP public/runtime schemas derive availability from the existing registry.

Findings for other supported roots, including repeated first-run objects in verification, are excluded from that run's batch contribution. **Possible child findings are preserved:** function modules, includes and other enclosing types outside the batch-root contract stay visible with their identity and worklist provenance. Their run remains incomplete and coverage counts unknown because parentage is unproven. This avoids reporting a container clean after silently dropping child findings, without introducing parent rollups or extra metadata requests. Original worklist totals and actual excluded counts remain available.

The implementation reuses the existing executor, with no cache, scheduler, automatic chunking or per-object retry loop. Legacy verification is skipped if too little time remains for settlement. Multi-target batches retain one selected target/identity and the existing read-only workload boundary. FUNC/INCL and package expansion remain outside the input-array contract; package calls remain available separately. Two worklists observe separate execution instants, so callers should keep source stable during the check.

Validation:

- **5,985 unit tests pass across 199 files**, including normal-concurrency CI runs on Node 22/24 and a local four-worker run. This follow-up adds 19 tests: seven failing before-fix cases (including a corrected existing expectation) reproduce child loss and TABL URI rejection. All nine targeted mutations were detected by assertions, covering independent cancellation guards, network recovery, returned Unicode identity, SRVB restrictions, child retention/counts, TABL resources and DEVC rejection.
- Default-concurrency local runs hit the unchanged CLI subprocess timeout also reported by the reviewer; its 27 tests pass in isolation and the entire suite passes with four workers, without changing tests or timeouts. Both Node CI jobs pass with the normal configuration.
- Typecheck, production build, Biome, policy validation, size/schema budgets and staged credential scan pass. No schema/snapshot or budget changes in this follow-up.
- Read-only live before/after checks on SAP_BASIS 758 preserve single/batch equivalence: the function group has zero findings, the program has the same 12 findings, and the table has a complete zero count. The final 20-class regression reports all 20 objects with the **same 150 findings** and a 73,774-byte response, retaining the earlier 11.6% payload reduction.
- The live container calls returned root-owned findings; separate child rows and alternate TABL source echoes are covered by constructed SAP responses. Prior 750/816, DDIC and lifecycle evidence is linked below. Successful live domain/data-element coverage under suitable variants and live BTP tenant verification remain unconfirmed.
- **All 16 GitHub checks pass on commit `321e777b`.** SAP integration: **226 passed / 87 skipped**; end-to-end: **136 passed / 7 skipped**, with zero failures. Cancelled SAP jobs were retried after the shared queue cleared; the successful final run is [CI attempt 2](https://github.com/arc-mcp/arc-1/actions/runs/34453232487/attempts/2).

Plans and review evidence:

- [Original plan](https://github.com/arc-mcp/arc-1/blob/321e777b06097251bd90f2dd53e265d38ea8bb0e/docs/plans/2026-09-09-atc-object-batches.md)
- [Initial verification](https://github.com/arc-mcp/arc-1/blob/321e777b06097251bd90f2dd53e265d38ea8bb0e/docs/research/2026-09-09-issue-770-implementation-verification.md)
- [DDIC verification](https://github.com/arc-mcp/arc-1/blob/321e777b06097251bd90f2dd53e265d38ea8bb0e/docs/research/2026-09-10-issue-770-ddic-verification.md)
- [First review: all 15 decisions](https://github.com/arc-mcp/arc-1/blob/321e777b06097251bd90f2dd53e265d38ea8bb0e/docs/research/2026-09-10-pr772-claude-review.md)
- [Second review plan](https://github.com/arc-mcp/arc-1/blob/321e777b06097251bd90f2dd53e265d38ea8bb0e/docs/plans/2026-09-10-atc-batch-rereview.md) and [assessment, before/after tests and final review](https://github.com/arc-mcp/arc-1/blob/321e777b06097251bd90f2dd53e265d38ea8bb0e/docs/research/2026-09-10-pr772-rereview.md)

Closes #770.


xXFracXx:
Thank you for this, the batching of up to 20 objects with the coverage verification is very helpful for us.

One request: our objects are DDIC (TABL, DTEL, DOMA), which are outside the initial batch types. Could you consider extending the batching to those as well?

PR #785, opened 2026-09-14
fix(query): fit freestyle SQL lines and honor minimal errors

MWKAnalytics:
## Goal

Preserve long freestyle SQL requests and honor `ARC1_MINIMAL_ERRORS` for classified SAPQuery errors.

## Changes

- Fit SQL into 255-character physical lines at the shared freestyle send function, covering SAPQuery, TABLE_QUERY, batches, and internal query callers.
- Wrap only between tokens, preserving quoted literals and comment boundaries. Refuse overlong input that cannot be split safely.
- Separate safe-break scanning from line wrapping, with explicit quote state and no sentinel values; add boundary regressions without introducing a parser dependency or changing accepted input.
- Keep the required minimal-error setting and single error formatter: minimal mode returns the status and ARC-1 hint without SAP diagnostics or paths.
- Remove the earlier statement-length hint, inaccurate tool-description sentence, redundant tests, and documents based on that premise.
- Keep existing IN-list chunking unchanged; its removal needs a separate compatibility review.

Live rechecks on SAP_BASIS 7.58 and 8.16 established that the limit applies to physical lines, not total statement length. The earlier space-padded 8.16 probes did not establish support for long ordinary one-line SQL.

## Validation

- 220 unit-test files / 6,817 tests pass.
- Typecheck, lint, build, docs build, policy validation, and file/schema-size checks pass (lint has three pre-existing informational diagnostics).
- Live 7.58 and 8.16 checks pass for ordinary long SQL, exact-256 input, CRLF, structured queries, a 600-value ordered IN-list, safe refusal, and both error-disclosure modes.
- Independent candidate review found no concrete remaining regression; 1,500 whitespace, quoting, and Unicode variations preserved SQL tokens.
- The readability follow-up matches the preceding implementation across 15,720 generated/boundary cases; a separate abaplint lexer check confirms unchanged tokens for 5,000 generated SQL cases. Added regressions cover doubled quotes around columns 254–256, inline-comment CRLF boundaries, and refusal when no advancing break exists.

The available 7.50 SP02 system has no working freestyle handler, so the reporter's SP23 environment remains unverified. CI test jobs are excluded from the current review at the maintainer's request because shared systems are running concurrent PR tests.


MWKAnalytics:
Deployed `main` (with this change) to our **ECC EhP8, SAP_BASIS 7.50 SP23** system and ran the three cases. The line-wrapping works on 7.50 too — the release you had not verified.

| Case | Result |
|---|---|
| Control, short single-line `SELECT` | executes normally |
| **350-character single-line `SELECT`** — 7 columns with long `AS` aliases, 3 `WHERE` predicates, on `TADIR` | **executes**, 635 rows reported, all aliases returned intact. Before this change the same statement failed with `Only one SELECT statement is allowed` |
| Single-line `SELECT` whose length sits inside one ~300-character quoted literal | refused client-side: `Cannot fit freestyle SQL into SAP's 255-character lines. Add line breaks between tokens; long literals, comments, and templates cannot be split safely.` |

The third case is the one I most wanted to see: ARC-1 refuses it itself, with a message that says what to do, instead of sending it and letting SAP return the grammar error that started this issue. The boundary behaves as designed.

So `FREESTYLE_SQL_LINE_MAX = 255` holds on 7.50 exactly as on 7.58/8.16, and the wrap makes the limit invisible for ordinary statements. If it is useful, the comment in `table-query.ts` (*"verified on 7.58 and 8.16"*) could be extended to 7.50.

Thanks for turning our issue into an actual fix rather than a better error message — this removes a daily annoyance for our consultants on ECC.

PR #801, opened 2026-09-17
fix(audit): authenticate the BTP Audit Log sink over mTLS and reject non-x509 bindings

### LINKED ISSUE #799: BTP Audit Log sink logs 'enabled' with a non-x509 binding and then fails silently — no event is ever written
## Summary

When the `auditlog` (plan `premium`) service instance is created without X.509 parameters, the binding carries `uaa.clientid` / `uaa.clientsecret` but no `uaa.certurl` / `uaa.certificate` / `uaa.key`. `parseBTPAuditLogConfig()` accepts that binding without validation, the startup log prints `INFO: BTP Audit Log sink enabled {"url": …}`, and every subsequent send fails inside `.catch()`. **No event is ever written, and nothing tells the operator.**

## Environment

- ARC-1 v1.2.0 on SAP BTP Cloud Foundry (eu10), five instances, each bound to an `auditlog` / `premium` instance created with a plain `cf create-service auditlog premium <name>` (no `-c` parameters) and bound with a plain `cf bind-service`.

## Reproduction

1. `cf create-service auditlog premium arc1-auditlog` (no parameters), `cf bind-service arc1-mcp-server arc1-auditlog`, `cf restart`.
2. Observe startup log: `INFO: BTP Audit Log sink enabled {"url":"https://api.auditlog.cf.eu10.hana.ondemand.com/premium"}`.
3. Run any `SAPQuery` / `SAPRead` tool call.
4. Read the subaccount's audit log through the Audit Log Retrieval API (`auditlog-management`, `GET /auditlog/v2/auditlogrecords?category=audit.data-access`).

## Observed

- Retrieval API: `204 No Content` for `audit.data-access` over 30 days; without a category filter, hundreds of platform `audit.security-events` records but **zero** records with `object.type = "MCP Tool Call"`, across all five instances, after a full day of tool calls.
- Binding credentials (field names only): `credentials.uaa` = `tenantmode, sburl, subaccountid, credential-type, clientid, xsappname, clientsecret, serviceInstanceId, url, uaadomain, verificationkey, apiurl, identityzone, identityzoneid, tenantid, zoneid` — no `certurl`, `certificate`, `key`.
- No error or warning about the sink in the application log.

## Where

- `src/server/sinks/btp-auditlog.ts` — `parseBTPAuditLogConfig()` selects the `premium`/`oauth2` binding and returns `{ url, uaa: { url, certurl, clientid, certificate, key } }` with whatever is present (possibly `undefined`), no validation.
- `fetchToken()` in the same file posts to `${this.config.uaa.certurl}/oauth/token` with `certificate`/`key`; with `certurl === undefined` this cannot succeed.
- `sendEvent(...).catch(...)` swallows the failure per event.

SAP's own documentation ("Audit Log Write API for Customers") states that the premium plan uses mTLS via X.509 and that the instance must be created with `xs-security.oauth2-configuration.credential-types: ["x509"]` and bound with `xsuaa.credential-type: "x509"` (plus `validity`), and that these certificates expire and must be rotated by re-binding.

## Expected

1. **Fail loudly**: if the selected binding lacks `certurl`/`certificate`/`key`, log an `ERROR` at startup naming the missing fields and the required `cf create-service … -c` / `cf bind-service … -c` parameters (or refuse to start when `SAP_AUDIT_REQUIRED`-style strictness is configured). The current `INFO: … sink enabled` line should not be printed in that case.
2. **Surface send failures**: at least a rate-limited `WARN` when audit posts fail, so a dark sink is visible in the app log.
3. **Docs**: `btp-cloud-foundry-deployment.md` / `security-guide.md` should show the X.509 create/bind parameters (also the MTA `requires.parameters.config.xsuaa.credential-type: x509` form) and mention certificate validity / rotation, since an expired binding reproduces exactly this silent state later.

Happy to contribute the validation + doc PR if wanted.

MWKAnalytics:
Fixes #799.

## Problem

The BTP Audit Log sink (`src/server/sinks/btp-auditlog.ts`) could not write a single event in 1.2.0, for three independent reasons that hid each other:

1. **The mTLS token request never presented the client certificate.** `getToken()` posted `grant_type=client_credentials&client_id=…` to `uaa.certurl` with the global `fetch`, which cannot carry a client certificate. The `*.authentication.cert.*` endpoint resets the TLS handshake without one; undici reports that as `TypeError: fetch failed`. The comment justifying this ("the CF buildpack handles certificate injection via `NODE_EXTRA_CA_CERTS`") does not hold — that variable adds trusted CAs, it never presents a client certificate.
2. **A non-x509 binding was accepted silently.** `parseBTPAuditLogConfig()` returned a config with `certurl`/`certificate`/`key` = `undefined` for the broker's default `binding-secret` binding, startup logged `INFO: BTP Audit Log sink enabled`, and every write failed inside `.catch()`.
3. **Data-access and data-modification records lacked the mandatory `data_subject`.** Once authentication worked, the Write API rejected tool-call records with `HTTP 400: 'data_subject' and 'data_subjects' properties cannot be both null or empty`.

## Fix

- `getToken()` requests the token through undici's `fetch` with an `Agent({ connect: { cert, key } })` built from the binding — the same undici-fetch rule `AdtHttpClient.doFetch` already follows (Node's built-in fetch does not accept npm-undici dispatchers). Handshake and HTTP failures now carry the endpoint and a certificate hint.
- `parseBTPAuditLogConfig()` throws `BTPAuditLogBindingError` when a premium/oauth2 binding lacks `certurl`, `certificate` or `key`; the message names the missing fields, the observed `credential-type`, and the exact `cf create-service` / `cf bind-service -c` parameters. `server.ts` logs that as an `ERROR` ("sink disabled") instead of the previous "optional" warning, and no longer prints "sink enabled".
- `buildPayload()` adds `data_subject` (`type: sap-system`, `role: data-owner`, `id.system` = target ?? destination ?? `configured-target`) to `data-accesses` and `data-modifications` records only; security events and configuration changes keep SAP's schema for those categories.
- Docs: `log-analysis.md` gains a provisioning section (x509 instance + binding parameters, the MTA `requires` form, certificate expiry/rotation, and the fastest verification via `cf logs … | grep BTPAuditLogSink`); `security-guide.md` and the CF runbook cross-reference it; `AGENTS.md` carries the one-line gotcha.

## Verification

- Unit: `tests/unit/server/sinks/btp-auditlog.test.ts` now mocks `undici` (fetch + Agent) instead of the global fetch; 26 tests cover the mTLS agent (cert/key from the binding, one agent per sink), the rejected-binding error (fields, plan, credential-type, cf parameters in the message), handshake/HTTP failure reporting, and `data_subject` presence/absence per category. Full `npm test`, `npm run typecheck`, `npm run lint` pass.
- **Live against a real subaccount (eu10):** the built sink from this branch, run with the `VCAP_SERVICES` of an x509 binding on a `auditlog`/`premium` instance, obtained a token over mTLS and received `201` for one event in each of the four categories (`data-accesses`, `data-modifications`, `configuration-changes`, `security-events`). The same binding with the 1.2.0 sink fails on every event with `fetch failed` (reproduced before the fix, see #799). The `data-accesses` record was subsequently found through the Audit Log Retrieval API with `object.type = "MCP Tool Call"`.
- Negative path reproduced live as well: a default `binding-secret` binding is what a plain `cf create-service auditlog premium …` + `cf bind-service` produces; with this change startup reports it as an error with the fix instructions.

## Notes for operators

- Instances and bindings must be created with x509 parameters (commands in `log-analysis.md`); `xsappname` must be unique per subaccount.
- Binding certificates expire after `validity` days — rotate by unbind/bind + restart, or the sink goes dark again (now visibly, via `[BTPAuditLogSink] Failed to write audit event: …`).
- `cf bind-service … -c` once returned `OK` while producing a `binding-secret` binding; verify the binding's `credential-type` after every bind.


marianfoo:
will be fixed in #802

PR #802, opened 2026-09-17
fix: restore BTP Audit Log mTLS delivery

### LINKED ISSUE #799: BTP Audit Log sink logs 'enabled' with a non-x509 binding and then fails silently — no event is ever written

marianfoo:
## Goal

Restore the optional BTP Audit Log premium sink and preserve pending audit writes during graceful shutdown.

Closes #799.

## Changes

- Validate the API URL and required X.509 fields before registering the sink; exclude client secrets from the token client.
- Use the existing @sap/xssec client for mTLS token acquisition and caching.
- Include a sap-system/data-owner subject on data-access and data-modification records, using the public target or a stable single-target fallback.
- Report delivery errors through rate-limited structured warnings without failing tool calls.
- Provide inactive-by-default MTA service and binding configuration; keep setup guidance concise and certificate rotation in the administration guide.
- On SIGTERM/SIGINT, stop transports, flush audit sinks, close the cache, and exit. A five-second deadline bounds stalled shutdowns; repeated signals do not bypass the flush.
- Wait for healthy sinks even if another fails, and for file appends already started by the periodic timer.
- Apply the maintainability review: remove unused fields/fallbacks, duplicate tests, the completed plan, and redundant documentation. Keep the small error-reporter injection so the sink stays independent of the global logger.

## Validation

- All 6,812 unit tests pass, including 12 shutdown regressions.
- Real-process SIGTERM/SIGINT tests hold an in-flight BTP audit post across shutdown, then verify delivery before exit; stdio/HTTP and cache/no-cache paths are covered. Temporarily removing the flush reproduces premature exit and makes the regressions fail.
- Typecheck, lint, build, policy validation, file/schema budgets, strict documentation build, and all shipped MTA descriptor validations pass.
- Earlier base and Audit-Log-enabled MTAR builds retained both X.509 parameter sets.
- The [reporter's eu10 rerun on 8f41a670](https://github.com/arc-mcp/arc-1/pull/802#issuecomment-5710239435) returned HTTP 201 for all four audit categories. Two records were already retrievable; the other two were still ingesting when the comment was posted. This was before the current simplification/shutdown follow-up.
- Shared CI test jobs are excluded from this assessment at the maintainer's request. No blocking review findings remain.

## Limits

Delivery remains best-effort: SIGKILL, shutdown timeouts, and service outages can lose pending records. The pre-existing SAP Converged Cloud endpoint-path difference is outside this fix.


MWKAnalytics:
Live-tested this branch (`6a87ec8e`) against a real `auditlog`/`premium` **x509** binding on eu10 — built locally, `VCAP_SERVICES` taken from the binding, one event per category through `BTPAuditLogSink.write()` + `flush()`:

| Category | Result |
|---|---|
| `security-events` (`auth_scope_denied`) | 201 |
| `configuration-changes` (`tool_call_start` SAPTransport) | 201 |
| `data-modifications` (`tool_call_end` SAPWrite) | 201 |
| `data-accesses` (`tool_call_start` SAPRead) | **400** — `{"subject":"'data_subject' and 'data_subjects' properties cannot be both null or empty"}` |

So the mTLS part is fixed — the `XsuaaService` token flow works against the certificate endpoint, and that is the better choice over a hand-rolled agent. But every ordinary tool call (`SAPRead`, `SAPSearch`, `SAPQuery` → `data-accesses`) is still rejected by the Write API's schema, which requires `data_subject` on data-access records (SAP Help, *Audit Log Write API for Customers* → "Audit Log Message Format" / "Writing a Data Access Event"). Without it the sink authenticates and then drops the bulk of the audit trail, one rate-limited warning per minute.

This is the third defect from #799; it only becomes visible once authentication works, which is presumably why it was never seen. In #801 we added, for `data-accesses` and `data-modifications` only:

```ts
data_subject: {
  type: 'sap-system',
  role: 'data-owner',
  id: { system: event.target ?? event.destination ?? 'configured-target' },
}
```

(the SAP system whose data the caller touches; security events and configuration changes keep SAP's schema for those categories). With that in place the same live run returns 201 for all four categories and the record is retrievable through the Audit Log Retrieval API with `object.type = "MCP Tool Call"`. Tests for presence/absence per category are in `tests/unit/server/sinks/btp-auditlog.test.ts` on that branch — feel free to lift them; happy to rebase our part onto this branch instead if you prefer, and to re-run the live probe once `data_subject` is in. #801 can then be closed as superseded.

One observation for the record: in this run the `data-modifications` endpoint accepted a `tool_call_end` record *without* `data_subject`, while `data-accesses` rejected the `tool_call_start`. I would not rely on that — SAP's examples carry `data_subject` for both — but it explains why a write-only smoke test could look green.

marianfoo:
Excellent catch — thank you for testing all four categories instead of stopping at the successful security-event smoke test.

I added the exact sap-system / data-owner subject shape from your live probe in 941d41b8. It is emitted only for data-accesses and data-modifications, with target → destination → configured-target precedence. The regression coverage checks all three identity paths and verifies that security/configuration payloads do not receive the field.

The focused tests, full 6,791-test suite, typecheck, lint, budgets, and strict docs build are green. I also updated the dossier and PR description to record this third masked defect and your eu10 evidence. A rerun against the new head would be valuable if convenient.

MWKAnalytics:
Re-ran the four-category live probe against `8f41a670` (includes `941d41b8`) with the same eu10 x509 binding: `data-accesses`, `data-modifications`, `configuration-changes`, `security-events` — **all four accepted (201)**. Two of the four records are already visible through the Audit Log Retrieval API (`object.type = "MCP Tool Call"`, `data_subject` populated as `sap-system`/`data-owner`); ingestion of the rest is still running as I write this.

From our side #802 is complete — mTLS, binding validation and the data-subject schema all hold against the real Write API. Thanks for the quick turnaround.

PR #800, opened 2026-09-17
fix: gate unsupported data-source lineage metadata on SAP 7.50

marianfoo:
Fixes #798.

With `SAP_BLOCKED_DATA_SOURCES` enabled on SAP_BASIS 7.50/7.51, queries against otherwise allowed tables reached an ADT table-source resource that does not exist on those releases. ARC-1 now returns `DATA_POLICY_UNAVAILABLE` before data execution. Direct blocks, unsupported SQL, and blocked aliases already visible in a CDS graph retain their more specific denials.

Loaded discovery prevents the known-missing request. In XSUAA principal-propagation deployments where the startup identity cannot read discovery, ARC-1 attempts the caller's canonical table-source read and classifies its `404` as the same unavailable-policy outcome. Other failures remain `DATA_LINEAGE_UNRESOLVED`; a `404` also remains unresolved when discovery explicitly advertised the resource. This avoids treating a missing object or authorization problem on a supported target as a release limitation.

The policy stays fail-closed because replacement objects exist on 7.50 and the legacy table representations omit that relationship. The implementation adds one discovery value and one narrowly scoped error classification; it adds no probe, cache, release parser, configuration, or fallback endpoint. Operator documentation, the [research dossier](docs/research/issues/798-data-source-blocklist-nw750.md), and the [completed plan](docs/plans/completed/2026-09-17-issue-798-data-source-policy-nw750.md) record the compatibility boundary.

Validation:

- Reproduced the contributor's PP sequence for `SAPQuery`, `TABLE_QUERY`, and `TABLE_CONTENTS`: unknown discovery, exact `TABL/DT` search result, canonical source `404`, and no data request.
- 416 focused tests passed, including all three paths, proactive discovery denial, graph/direct denial precedence, advertised-resource `404`, minimal errors, and audit behavior.
- All 6,798 unit tests across 220 files passed.
- Typecheck, lint, policy validation, production build, strict docs build, file-size/schema budgets, and diff checks passed.
- Earlier live verification passed on SAP_BASIS 750 and 758. Shared SAP CI integration/e2e jobs remain excluded from the merge recommendation at the maintainer's request because multiple PRs are testing concurrently.

Issue #798, opened 2026-09-17
SAP_BLOCKED_DATA_SOURCES denies every data request on SAP_BASIS 7.50 — lineage resolver reads /ddic/tables/<name>/source/main (absent before 7.52), no release gate


MWKAnalytics:
## Summary

With a non-empty `SAP_BLOCKED_DATA_SOURCES`, **every** data request on a SAP_BASIS 7.50 system is denied with `DATA_LINEAGE_UNRESOLVED` — including tables that are not on the list, and including `SAPRead(TABLE_QUERY)` / `SAPRead(TABLE_CONTENTS)`. The lineage resolver reads `/sap/bc/adt/ddic/tables/<NAME>/source/main` for every direct source; that resource does not exist before SAP_BASIS 7.52, SAP answers `404`, and the policy fails closed as designed.

The blocklist itself behaves exactly as documented (direct and in-join matches are denied with zero SAP calls). The gap is that the feature has no minimum-release statement and no release gate, although `tools.md` → "Release gates for pre-7.52 systems" already records that `/sap/bc/adt/ddic/tables` is absent on NW 7.50 and gates `SAPWrite` on it with a clear hint.

## Environment

- ARC-1 v1.2.0 on SAP BTP Cloud Foundry, XSUAA + principal propagation, `SAP_ALLOW_DATA_PREVIEW=true`, `SAP_ALLOW_FREE_SQL=true`
- Target: ECC EhP8, **SAP_BASIS 7.50 SP23**
- `SAP_BLOCKED_DATA_SOURCES` = 21 exact table names (validated with `parseBlockedDataSourcesCsv`, startup log shows `blockedDataSourcesCount: 21`)

## Reproduction

1. Set `SAP_BLOCKED_DATA_SOURCES=USR02,PA0002` (any non-empty, valid list) against a 7.50 system.
2. `SAPQuery` with `SELECT COUNT(*) AS N FROM T000` (or `TADIR` — any plain transparent table not on the list).
3. Also `SAPRead(type="TABLE_QUERY", name="T000", columns=["MANDT"])` and `SAPRead(type="TABLE_CONTENTS", name="T000")`.

## Observed

All three are denied:

```
DATA_LINEAGE_UNRESOLVED: the request was denied by the administrator's data-source policy before any SAP data request was executed (executed=false, decisionId=dsp_df29096baea2c98a) …
```

Server log (`data_source_policy_decision`):

```
[...]

PR #833, opened 2026-09-21
feat(policy): experimental sensitive data-source list with audited justification

MWKAnalytics:
## Change

- Problem and resulting behavior: Operators who approve data access for developers had only a hard
  denial (`SAP_BLOCKED_DATA_SOURCES`) to govern sensitive tables. This adds
  `SAP_SENSITIVE_DATA_SOURCES` / `--sensitive-data-sources`: an exact-name list whose *direct* reads on
  `SAPQuery`, `TABLE_QUERY` and `TABLE_CONTENTS` pause with a new stable code `DATA_SOURCE_SENSITIVE`
  (executed=false, decision id) until the caller repeats the request with a `justification`. The
  justification, the matched sources and the calling user are recorded in the
  `data_source_policy_decision` audit event and the request runs.
- Release coverage: release-independent by construction. The check needs no ADT metadata, so it is
  currently the only data-source control ARC-1 can enforce on SAP_BASIS 7.50/7.51, where the blocklist
  correctly reports `DATA_POLICY_UNAVAILABLE` (#798, #800). The two are complements: the blocklist
  proves transitive lineage and denies; this list checks direct names and records. A source on both
  lists stays blocked; startup warns about the overlap.
- Related: #740 (data-source policy), #798/#800 (blocklist on 7.50); adjacent to roadmap FEAT-66.
- Before → after: `SAPQuery(sql="SELECT MANDT FROM T000")` with `T000` listed →
  `DATA_SOURCE_SENSITIVE: … repeat the same request with the justification parameter …`; retry with
  `justification="…"` → rows, audit event with `sensitiveSources=["T000"]`, `justification`, `user`.

## Compatibility and scope

- Default-off; the empty value keeps today's behavior byte-for-byte. Shipped descriptors carry
  `SAP_SENSITIVE_DATA_SOURCES=""` like the blocklist for the one-field rollback.
- New optional tool parameter `justification` on SAPQuery and SAPRead (JSON schema, Zod, handlers,
  fixtures regenerated; schema-budget ratchets bumped by the measured delta).
- This is an attestation control, not a boundary: any caller can supply a justification, so the list
  documents who read what and why. Documented as such; the blocklist remains the denial. ARC-1 cannot
  tell a person's answer from the model's over the tool channel (noted at FEAT-66).
- BTP Audit Log sink: the sensitive pause and the justified allow are forwarded to `security-events`;
  other policy decisions are unchanged (server log only).
- `TABLE_CONTENTS` with `sqlFilter` is refused while either list is active (a condition can carry a
  subquery), same as the blocklist today.

### Automated checks

- `npm run lint`, `npm run typecheck`, `npm run check:sizes`: pass.
- `npm test`: 7058 passed; 12 failures are pre-existing Windows-only tests (POSIX file permissions,
  SIGTERM/SIGINT, absolute-path fixtures) that fail identically on unmodified `main`.
- New tests: policy unit (normalization, exact match, minimal message), client (all three data paths,
  justified allow with audit fields, blocklist precedence, sqlFilter refusal, lineage still evaluated
  when both lists are active), dispatch (client message, minimal mode, schema acceptance and 500-char
  cap, errorClass), config, safety, effective-policy log, BTP sink categorization, UI/config show.

### Real-system testing

- Status: Tested.
- Tested build and environment: this branch at 3b2f46b6 (on main 4253d00b), pushed onto the existing
  BTP CF app (Node buildpack 1.9.4, XSUAA OAuth proxy, principal propagation, `ARC1_MINIMAL_ERRORS=true`)
  against ECC EhP8, SAP_BASIS 7.50 SP23; client Claude (hosted connector), 2026-09-21 20:57–20:59Z.
  `SAP_SENSITIVE_DATA_SOURCES=T000` set via `cf set-env` for the test only.
- Scenario and observed result:
  - `SAPQuery(sql="SELECT MANDT FROM T000")` → `DATA_SOURCE_SENSITIVE`, minimal message without the
    table name, decisionId `dsp_230c33b6109ed8d6`; audit event `deny`, `metadataRequests: 0`.
  - Same statement with `justification` → 3 rows; audit event `allow` with `sensitiveSources=["T000"]`,
    the justification and the user; `metadataRequests: 0`, one freestyle POST.
  - `SAPRead(type="TABLE_QUERY", name="T000", columns=[…])` without / with `justification` → same
    pause (`dsp_0e0b232d9a90d9d5`) and same allow (`dsp_b90320cc7aa680f8`).
  - Control `SAPQuery(sql="SELECT BUKRS FROM T001")` without justification → 2 rows, ordinary allow
    without sensitive fields.
  - BTP Audit Log retrieval (`category=audit.security-events`): all four sensitive decisions present
    with user and justification text; the T001 allow correctly absent.
  - Startup log: `sensitiveDataSources=1/<fingerprint>`, no contradiction warning. Rolled back to the
    previous revision afterwards; environment verified clean.
- Limits and follow-up verification: not tested on 7.58; multi-target mode covered by unit tests only;
  not tested with an active blocklist on a live system (unit-covered: blocklist wins, lineage still runs).

## Roadmap

- [x] Checked the idea roadmap. FEAT-66: added an "Adjacent, shipped" note naming this feature and the
  remaining gap (a confirmation that provably reaches a person). No other item affected.

Created with support of Claude Code.


marianfoo:
nice idea, but why having this in the first place? yes this will be recorded but i dont like a kind of llm gated source, if the user know what he needs to say he can access it? I´d rather go for hard decisions then LLM gated things.
sorry, but i will not implement this

MWKAnalytics:
**Follow-up idea (not part of this PR): a confirmation that provably reaches a person**

This PR deliberately stops at attestation: the model supplies the `justification`, and ARC-1
cannot tell over the tool channel whether a person was actually asked. For sensitive tables
some operators will want the human in the loop, and FEAT-66's elicitation route is not
available to them today: Claude Code does not deliver elicitation to remote Streamable HTTP
servers (anthropics/claude-code#85442), Claude Desktop has no elicitation at all, and
Anthropic's `_meta["anthropic/requiresUserInteraction"]` tool annotation is honoured only by
Claude Code.

The client-agnostic option is an out-of-band consent step, the same pattern as OAuth consent
or the device-code flow. It needs nothing from the MCP client beyond showing a URL in a tool
result, so it behaves identically in Claude Code, Claude Desktop, Cursor, Codex, Copilot or
any future client, and it does not depend on which client honours which annotation:

1. `DATA_SOURCE_SENSITIVE` carries a one-time consent URL (`/consent/<signed-id>`) instead of, or
   next to, the justification retry.
2. The person opens it in a browser, signs in through the existing XSUAA flow, sees system,
   user, direct sources and the statement, **types the business reason themselves** and
   confirms or declines.
3. ARC-1 checks that the browser identity equals the MCP user, stores the confirmation bound to
   that request (single use, ~5 min), and audits it with the user from the browser session.
4. The model repeats the request with the same request id; the server finds the confirmation
   and runs it. The tool channel only ever carries an id, so nothing the model can produce
   substitutes for the click.

This closes the gap this PR leaves open on purpose. With `justification` alone, nothing stops
an agent from writing a plausible reason itself and retrying without ever asking anyone; the
audit trail would then record the agent's words under the user's name. With the consent page,
both the reason and the confirmation originate in the person's own browser session, and the
agent never authors either of them. The `justification` parameter would stay for the
attestation-only mode; the consent flow simply supersedes it where an operator turns it on.

Fail-closed by construction (no confirmation → the pause stays), a complement to the ceiling
(the blocklist still wins), and it reuses `@arc-mcp/xsuaa-auth` (authorization-code flow,
signed state, callback). Open design points: in-process confirmation store vs. a signed
variant for multi-instance deployments, an optional forced re-login (`max_age=0` / MFA) so an
agent driving the user's own browser cannot click for them, and how it should eventually hand
over to elicitation once FEAT-66's triggers fire.

Before we build it: would you consider this direction for ARC-1, or should it stay a
deployment-specific extension? Happy to write it up as an issue with the full design if useful.

marianfoo:
also this, not a fan, too complicated to setup and maintain and i dont see the value for this

MWKAnalytics:
@marianfoo Understood, and thanks for the quick look. Fair point: a control whose outcome depends on what
the caller types is only as strong as the caller, and that is a class you don't want in ARC-1.
We accept that, we won't pursue the follow-up idea, and future requests from our side will stick
to hard gates.

One note on where it came from, for context only: our concern was agents running in auto mode
that read sensitive tables nobody meant to expose, because a loosely worded developer request
("find out why this customer's invoice is wrong") legitimately leads the model into KNA1 or
PA0002 without the developer being aware of it. The pause was meant to make that moment
visible and to leave a record with a stated purpose, which is what GDPR accountability
(Art. 5(2), records of processing) and the EU AI Act's emphasis on human oversight and
traceability push operators towards. A hard denial covers "never", not "only when someone
consciously decided to" — that middle ground is what we were after, and it may simply be a
deployment-side concern rather than ARC-1's.

PR #835, opened 2026-09-22
feat(dumps): window and page the ST22 feed past SAP's 100-entry ceiling

marianfoo:
## Change

- **Problem and resulting behavior:** SAP serves at most **100 entries** from `/sap/bc/adt/runtime/dumps` and silently ignores `$skip`, so ARC-1's `maxResults` cap of 200 was unreachable fiction and every broader question was answered from a truncated list. On the 2025 test system, **762 of 862 dumps were invisible to any single call**. `listDumps` now walks the inclusive `to` bound (keying by dump id to drop the repeated boundary entry), and `SAPDiagnose(action="dumps")` accepts `from`/`to` — both already existed in the Zod and JSON schemas; only this handler ignored them.
- **Related issue(s):** none.
- **Reproducer (before → after)** on a4h-2025 (SAP_BASIS 816), which holds 862 dumps:

  ```
  SAPDiagnose(action="dumps", maxResults=500)
  before -> 100 dumps (silently capped by SAP; $top=9999 also returns 100)
  after  -> 500 dumps, 500 distinct, span 2026-06-25 .. 2026-09-21, newest-first

  SAPDiagnose(action="dumps", from="2026-09-15", to="2026-09-20T23:59:59Z")
  before -> from/to ignored by the handler; whole feed returned
  after  -> 34 dumps, all inside the window
  ```

### Contract evidence (raw ADT, live)

| Param | Behavior on 758 and 816 |
|---|---|
| `$top=N` | Honored downward only. `$top=101/150/200/500/9999` all return exactly **100**. |
| `$skip=N` | **Silently ignored** — `$top=5&$skip=5` returns the identical newest 5. |
| `from`/`to` | **Honored**, inclusive on both ends (verified against an exact `atom:published` value, ±1s). |
| `$query=and(equals(user,X))` | Honored; the `and(...)` wrapper is mandatory (bare `equals(...)` returns 0). |
| malformed bound | **Returns the whole feed** — hence the fail-closed validation in this PR. |

Also probed and rejected as ceiling-lifters: `maxResults`, `$maxrows`, `rows`, `count`, `numberOfResults`, `$count`, `$inlinecount`, `size` — all still 100.

## Compatibility and scope

- The default path still issues **exactly one request** (`maxResults` 50 ≤ one page), so existing callers are unchanged.
- `maxResults` for dumps is now capped at 500 (paged); the other feeds keep their 200 single-shot cap.
- Time bounds normalize to SAP's `YYYYMMDDHHMMSS` and accept `YYYY-MM-DD`, ISO with fractional seconds (`toISOString()`), and UTC offsets. An unparsable bound is **rejected** rather than passed through, because SAP answers a bound it cannot parse with the entire feed.
- No new action, scope, or safety surface: `dumps` stays a read, and the multi-target allowlist derives from the same definitions.
- `src/adt/diagnostics.ts` line budget raised 1845 → 1902, annotated in `check-file-sizes.mjs`.

### Automated checks

- `npm test` → 232 files, **7088 passed**
- `npm run typecheck`, `npm run lint`, `npm run build`, `npm run validate:policy` (128 entries / 14 schemas), `npm run check:sizes` → all pass
- Tool-definition fixtures regenerated; the diff is description text only, no structural change
- New unit tests: multi-page cursor assembly with a duplicated boundary entry, single-request short-page stop, `from`/`to` normalization, `toISOString()` fractional seconds, UTC-offset conversion, and fail-closed rejection

### Real-system testing

- **Status:** Tested (on-prem, two releases; a third had no dumps to exercise)
- **Tested build and environment:** this branch at `8103348c`, built `dist/` driven via `arc1-cli call SAPDiagnose`; a4h 2023 (S/4HANA 2023, SAP_BASIS **758**) and a4h-2025 (ABAP Platform 2025, SAP_BASIS **816**), client 001, on-prem Basic auth over HTTPS. Raw-ADT probes used direct `curl`/`fetch` against the same systems.
- **Scenario and observed result:**
  - `{"action":"dumps"}` → 758: 33 dumps; 816: 50 (default), one request each
  - `{"action":"dumps","maxResults":500}` → **816: 500 dumps, 500 distinct ids, newest-first order verified**, span 2026-06-25 .. 2026-09-21; 758: 33 (all it has), clean termination
  - `{"action":"dumps","from":"2026-09-15","to":"2026-09-20T23:59:59Z"}` → 816: 34, 758: 6 — every entry inside the window
  - `{"action":"dumps","user":"MARIAN","maxResults":200}` → 816: 200 dumps, all owned by MARIAN (filter carried across pages); 758: 28
  - `{"action":"dumps","from":"yesterday"}` → rejected with `Invalid from timestamp "yesterday". Use YYYYMMDDHHMMSS, YYYY-MM-DD, or an ISO timestamp such as 2026-09-15T00:00:00Z.` (before this PR SAP would have returned the entire feed)
  - A standalone cursor dry-run assembled **862 distinct dumps in 10 requests** back to 2022, terminating on its own.
- **Limits and follow-up verification:** NPL (750) has no dumps recorded, so the paging path is unexercised there; the single-page path is release-independent. Not tested on BTP/Steampunk (no ST22 dump corpus available) or via HTTP/XSUAA transport — this change is transport-agnostic. Read-only throughout; no writes, locks, or transports involved.

## Roadmap

- [x] Checked the [idea roadmap](https://github.com/arc-mcp/arc-1/blob/main/docs_page/roadmap.md) and updated it where needed.

No roadmap impact — the roadmap has no dump/ST22 entry, and this change is fully implemented here.


marianfoo:
Addressed all three findings in `f0ec58bb`.

**1. [P2] Pagination silently truncating on a stalled cursor — fixed.**

Confirmed, and the probe you suggested turned out not to resolve it. I first tried disambiguating with a one-second-earlier lookahead, but once ≥100 dumps share the cursor second, *both* branches are lossy: with nothing older, SAP still only hands back 100 of however many sit in that second, and it reports no total. Completeness is unknowable either way, so the honest answer is to stop rather than pick a story:

```
Cannot page past 20260901000000: at least 100 dumps share that second, which is SAP's
maximum per request, so older ones are unreachable. Narrow the query with user, or read
that second with to=20260901000000.
```

Only a caller asking for more than one page can reach this — with `maxResults <= 100` the loop exits after the first request, so a single-page read from inside a burst still returns its 100 normally. Both are covered by regression tests driven by a mock feed that behaves like SAP (newest-first, inclusive `to`, hard 100 cap): your exact 150-same-second + 50-older repro now throws, and `maxResults=100` against the same corpus returns 100 in one request.

**2. [P2] Impossible dates — fixed, and it was worse than reported.**

They did not merely reach HTTP. `2026-02-30T00:00:00Z` is not rejected by V8 — it **rolls over to 2026-03-02**, so a typo returned *another day's dumps as a success*, which is the exact failure mode the validation existed to prevent. Bounds now round-trip through `Date` and must serialize back to the same stamp. A trailing UTC offset is split off and applied *after* that check, so `2026-02-30T00:00:00+02:00` is rejected too rather than bypassing it via the fallback. Fractional seconds and valid offsets still pass. Rejection cases are table-tested: month 13, Feb 30, hour 25, and Feb 30 behind an offset.

**3. [P3] authorization_trace cap — fixed.** Confirmed `clampPreviewRows` → `MAX_TABLE_QUERY_ROWS = 10_000`. Now reads: `Result limit (default 50; authorization_trace 100). Caps: dumps 500, system_messages/gateway_errors 200, authorization_trace 10000.` Snapshots regenerated.

**Validation.** `npm test` 7094 passed, plus typecheck, lint, `validate:policy`, `check:sizes`, `build`. Re-verified live on a4h 2023 (758) and a4h-2025 (816) against the rebuilt `dist/`: `maxResults=500` → 500 distinct dumps spanning 2026-06-25 .. 2026-09-22 on 816, windows filter correctly, the user filter still survives paging (200/200 MARIAN), junk bounds rejected. The burst path is mock-only — no system had ≥100 dumps in one second to reproduce it live (peak real clustering across 863 dumps on 816 is **2**).

**Known unrelated failure:** `tests/unit/server/oauth-redirect-policy.test.ts` ("round trips a supported manual callback") fails roughly 1 run in 7 on this branch and on a clean tree; it touches none of these files. Tracked separately, not fixed here.

Diagnostics line budget is now 1932 (+87), annotated in `check-file-sizes.mjs`.

marianfoo:
Confirmed and fixed in `e7f755e9`. You were right, and it was a regression I introduced: replacing the plain `Date` fallback with explicit offset handling is exactly what let these through — `Date` rejected both.

Reproduced against the review commit before fixing:

| input | before | plain `Date()` |
|---|---|---|
| `2026-09-15T08:00:00+02:60` | `from=20260915050000` | INVALID |
| `2026-09-15T08:00:00+24:00` | `from=20260914080000` (previous day) | INVALID |

The offset is now range-checked where it is split off, by constraining the match itself to RFC 3339 bounds (`hh` 00–23, `mm` 00–59):

```ts
const offset = raw.match(/([+-])([01]\d|2[0-3]):?([0-5]\d)$/);
```

A malformed offset then fails to match, so the remainder still carries `+`/`-` into the 14-digit check and is rejected before any HTTP call — same path and message as the other bad bounds, no separate branch.

Regression tests as requested, both bounds and both directions:
- rejection table extended with `+02:60` and `+24:00`, and every case in it now runs against **both `from` and `to`**, asserting the matching `Invalid <field>` message and `expect(http.get).not.toHaveBeenCalled()`
- conversion table covers a **positive** offset (`+02:00` → `20260915060000`), a **negative** one (`-05:30` → `20260915133000`), and the colon-less form (`+0200`), all matching what `Date()` produces for the same input

`+23:45` stays accepted, unchanged and consistent with `Date()`.

**Validation:** `npm test` 7098 passed (232 files), typecheck, lint, `validate:policy`, `check:sizes`, `build` all green. Re-verified live on a4h 2023 (758) and a4h-2025 (816) against the rebuilt `dist/` — `maxResults=500` → 500 distinct dumps, windows and the paged user filter unchanged. Diagnostics budget 1933 (+88).

The `oauth-redirect-policy.test.ts` flake remains the only known failure and is unrelated; it is being fixed separately.

marianfoo:
## Final cross-check against other implementations and SAP's own source

Checked what other ADT clients do and what SAP itself declares. **No functional gap found — and SAP's implementation confirms the design.** One genuine capability gap is parked as roadmap FEAT-74.

### SAP self-describes this feed, and it matches what was measured

`GET /sap/bc/adt/feeds` carries a `feed:extendedData` block for the dumps feed:

```xml
<feed:paging size="50" max="100"/>
```

The 100-entry ceiling established by probing is SAP's own **declared maximum**, and the default page size is 50 — the same default this PR keeps.

### SAP's handler confirms the paging contract

Read live from `CL_RABAX_ADT_RES_DUMPS` on a4h (758):

```abap
constants c_max_paging type i value 100 ##NO_TEXT.
...
if me->page_size is initial or me->page_size gt c_max_paging.
  me->page_size = c_max_paging.        " $top only narrows; 0 and >100 both clamp TO 100
endif.
...
me->page_size = me->page_size + 1.     " because (TODO at least) one dump is retransmitted!
...
if me->page_size > 0 and me->page_size < lines( l_snap_keys ).
  delete l_snap_keys from ( me->page_size + 1 ).
  l_add_next_link_for_paging = 'X'.
endif.
```

Three things this settles:

1. **The repeated boundary entry is by design**, not an artifact of reconstructing the cursor — SAP deliberately selects one extra row *because one dump is retransmitted*. Keying by dump id is the intended handling.
2. **A short page is exactly SAP's own terminator.** The next link is added only when more rows existed than the page size, so "fewer than `pageSize` returned" ≡ "no `rel="next"`". The loop stops at the same point SAP would.
3. `$top=0` clamps *up* to 100 (`is initial`), so it never means "unlimited".

### Why not follow the feed's `rel="next"` link

The feed does publish one, and it uses `to=<oldest entry>` — the same cursor this PR builds. Walked both to exhaustion on 816: **865 distinct dumps, 9 pages, identical sets, zero difference in either direction.**

Following it verbatim would be worse, because SAP *appends* rather than replaces the query string on each hop:

```
page 2: ?$top=100&sap-client=001&sap-client=001&to=20260731083406
page 3: ?$top=100&sap-client=001&sap-client=001&sap-client=001&to=...
```

The parameter accumulates every hop. Rebuilding the cursor avoids that while producing the same result.

### How other clients handle it

| Client | Approach |
|---|---|
| `marcellourbani/abap-adt-api` | `$query` only — no `from`/`to`, no `$top`, no paging |
| `marcellourbani/sapdumpmcp` | unfiltered `dumps()` |
| `vscode_abap_remote_fs` | unfiltered |
| `oisee/vibing-steampunk` | `from`/`to`, capped at 100, **no paging**; filters client-side |
| `yzonur/sap-adt-mcp` | client-side filtering |

Most stop at the 100-entry wall. None of the surveyed clients both page past it and validate bounds.

### Genuine gap found → roadmap FEAT-74 (`6b24ba74`)

The feed declares **11 filterable attributes** — `runtimeError`, `exception`, `objectName`, `package`, `packageHierarchy`, `component`, `responsible`, `objectResponsible`, `packageResponsible`, `user`, `datetime` — with `equals`/`notEquals`/`contains`/`notContains`/`greater`/`less`/`between`/`notBetween` and `queryDepth` 2. ARC-1 exposes only `user`. Verified honored on 816: `and(contains(objectName,SAPM))` → 11 entries, `and(between(datetime,20260901000000,20260910000000))` → 49.

Server-side filtering would beat paging-and-discarding for questions like "every TIME_OUT dump in package Z\*". Out of scope here — `from`/`to` covers the time-window case this PR is about — so it is parked as FEAT-74 (P3/S/Ready) rather than bolted on.

No `/sap/bc/adt/vit/runtime/dumps` REST alternative exists (it is a SAP GUI deep link, `type="application/vnd.sap.adt.sapgui"`, and 404s as a feed), and no uncapped route or total-count facility turned up.

